PR #23, opened 2021-05-31
just sum smol improvements

TopiSenpai:


api/command.go
```diff
@@ -30,6 +30,16 @@ func (c Command) FromGuild() bool {
 	return c.GuildID == nil
 }
 
+// ToCreate return the CommandCreate for this Command
+func (c *Command) ToCreate() *CommandCreate {
```
EveCodes31, line 34: What exactly is this for?

TopiSenpai: I actually use it here <https://github.com/DisgoOrg/disgommand/blob/5b1a6023218f3fe946bdb3628514f28c07249a24/command.go#L29>

EveCodes31: Ah, for getting the struct ready to create the command 👍

api/restclient.go
```diff
@@ -6,14 +6,8 @@ import (
...
-// Errors when connecting to discord
-var (
-	ErrBadGateway                 = errors.New("bad gateway could not reach discord")
-	ErrUnauthorized               = errors.New("not authorized for this endpoint")
-	ErrBadRequest                 = errors.New("bad request please check your request")
-	ErrRatelimited                = errors.New("too many requests")
-	ErrTooMuchCommands = errors.New("you can provide a max of 100 application commands")
-)
+// ErrTooMuchCommands returned if a Guild reached max of 100 Command(s)
+var ErrTooMuchCommands = errors.New("you can provide a max of 100 application commands")
```
EveCodes31, line 10: This should be ErrToManyCommands

EveCodes31 (changes requested):

api/restclient.go
TopiSenpai: ErrMaxCommands would be even better I guess

EveCodes31: That works

EveCodes31 (approved):

PR #49, opened 2021-07-06
make events pointer & not fire generic events

TopiSenpai:
* pointers for events should be better I hope

* I don't think there is any need to fire generic events as they can't be casted to their other events
at this point they are super useless and just require cpu time

* refactored some stuff in command event


EveCodes31 (approved):
no more generic events! 🎉

TopiSenpai:
I mean they are nice
but we can't cast them so they suck ;-;

PR #72, opened 2021-08-19
ratelimiter implementation

TopiSenpai:


TopiSenpai:
linting is not important here for now :p

rest/http_client.go
```diff
@@ -102,11 +128,17 @@ func (c *HTTPClientImpl) DoWithHeaders(ctx context.Context, compiledRoute *route
 		return NewError(rs, err)
 	}
 
+	// catch rate limit errors here
+	// TODO: we should retry on hitting 429
```
arynxd, line 132: unless the API can incorrectly 429, no we should just wait for retry-after

rest/http_client.go
```diff
@@ -20,21 +21,38 @@ type HTTPClient interface {
...
-	return &HTTPClientImpl{userAgent: userAgent, httpClient: httpClient, logger: logger}
+func NewHTTPClient(logger log.Logger, httpClient *http.Client, rateLimiter rate.RateLimiter, headers http.Header, userAgent string) HTTPClient {
+	if logger == nil {
+		logger = log.Default()
+	}
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+	if rateLimiter == nil {
+		rateLimiter = rate.NewRateLimiter(logger)
```
arynxd, line 39: would it be worth having a RateLimiterConfig to configure the behaviour of the ratelimiter? eg; handling retries and what should happen on a 429>

arynxd (changes requested):
submitted as request changes so we can discuss this before merging, no changes required yet

arynxd:
should we also have a ratelimited event? so users can provide custom logging impls for ratelimits. this event could be fired when we avoid a ratelimit aswell as when we actually hit 429

rest/http_client.go
TopiSenpai: My thought was putting that into the http client as that also handles the retry itself
But yes we should

rest/http_client.go
TopiSenpai: Yup 
This was meant with the correct delay
Just need to grab the bucket again in a loop over retry count

Skye-31:
> should we also have a ratelimited event? so users can provide custom logging impls for ratelimits. this event could be fired when we avoid a ratelimit aswell as when we actually hit 429

I think a separate event for each would be better, something like RatelimitPrevented and RatelimitReceived

rest/http_client.go
TopiSenpai: nvm I ended up putting it in the RateLimiter Config haha

TopiSenpai (commented):
I'll implement the event later

PR #73, opened 2021-08-23
request opts

TopiSenpai:
this aims to make it possible to add modifiers to requests like `context.Context`, `X-Audit-Log-Reason` and possible other stuff in the future

Additionally usecases:
* Checks
* Request delay


TopiSenpai:
this should be ready to review @Skye-31 
just ignore missing docs ty :D

_examples/webhook/example.go
```diff
@@ -2,29 +2,59 @@ package main
...
+	<-s
+}
 
-	cancel()
+func send(ctx context.Context, client webhook.Client, i int) {
+	_, err := client.CreateMessage(webhook.NewMessageCreateBuilder().SetContentf("test %d", i).Build(),
+		rest.WithCtx(ctx),
+		rest.WithReason("this adds a reason header"),
+		rest.WithDelay(2*time.Second),
+		rest.WithCheck(func() bool {
```
Skye-31, line 53: What's the purpose of this?

core/entity_builder_impl.go
```diff
@@ -278,34 +278,6 @@ func (b *EntityBuilderImpl) CreateGuild(guild discord.Guild, updateCache CacheSt
...
-	for _, emote := range guild.Emojis {
-		b.CreateEmoji(guild.ID, emote, updateCache)
-	}
-
-	// TODO: presence
-	/*for i := range fullGuild.Presences {
-		presence := fullGuild.Presences[i]
-		presence.Disgo = disgo
-		b.Disgo().Cache().CachePresence(presence)
-	}*/
```
Skye-31, line 308: Was all of this logic meant to be removed?

discord/gateway_intents.go
```diff
@@ -42,6 +42,9 @@ const (
 	GatewayIntentsAll = GatewayIntentsNonPrivileged |
 		GatewayIntentsPrivileged
 
+	GatewayIntentsDefault = GatewayIntentGuilds |
```
Skye-31, line 45: I don't think it's good practice to provide defaults for Intents, it's something that should be explicitly set by the user.

Skye-31:
Left some questions

core/entity_builder_impl.go
TopiSenpai: yep I moved it into the gateway event handler

discord/gateway_intents.go
TopiSenpai: then default would be `0`
idk if thats better

_examples/webhook/example.go
TopiSenpai: the checks are executed after the `delay` & `ratelimit` was awaited if necessary

those can be used to easily do last minute checks

discord/gateway_intents.go
Skye-31: I disagree, imo it should throw an error if not provided

core/entity_builder_impl.go
Skye-31: Ah, ok

discord/gateway_intents.go
TopiSenpai: that would require it to be a pointer which is ugly >.>

Skye-31: then defaulting to 0 is better

TopiSenpai: oki

PR #79, opened 2021-09-15
sharding support

TopiSenpai:


TopiSenpai:
this is pretty much ready for review/discussion

with the current system you always have a `sharding.ShardManager`

I can think of having either of both on the Bot struct

need ur input on that @Skye-31 :)

TopiSenpai:
The `core.Bot` struct is missing some methods which use `context.Context`

Skye-31:
> this is pretty much ready for review/discussion
> 
> with the current system you always have a `sharding.ShardManager` instead of `gateway.Gateway`
> 
> I can think of having either of both on the Bot struct
> 
> methods using the gateway would be checking which one is used ofc
> 
> need ur input on that @Skye-31 :)

I'd say allow both, and if a sharding manager isn't specified, provide a reasonable default

Skye-31 (approved):

Skye-31:
Could also be nice to have an option for the time delay between spawning each shard, and the maximum amount of time to wait when a shard is launching before throwing an error

TopiSenpai:
> Could also be nice to have an option for the time delay between spawning each shard, and the maximum amount of time to wait when a shard is launching before throwing an error

for starting a single shard it takes an context & I can add that time to the ratelimiter config 👍🏻

TopiSenpai:
> Could also be nice to have an option for the time delay between spawning each shard

done that :)

TopiSenpai:
what do you think of the init flow via the opts? @Skye-31

Skye-31:
lgtm

PR #84, opened 2021-09-27
better handling of ws closes

TopiSenpai:


TopiSenpai:
this needs some more testing

discord/errors.go
```diff
@@ -10,6 +10,7 @@ var (
 	ErrNoGuildMembersIntent    = errors.New("this operation requires the GUILD_MEMBERS intent")
 	ErrNoShardManager          = errors.New("no shard manager configured")
 	ErrNoGateway               = errors.New("no gateway configured")
+	ErrGatewayAlreadyConnected = errors.New("no gateway configured")
```
Skye-31, line 13: this is the wrong error message

Skye-31 (changes requested):

gateway/gateway.go
```diff
@@ -12,10 +12,10 @@ import (
 // Status is the state that the client is currently in
 type Status int
 
-// IsConnected returns whether you can send payloads to the Gateway
+// IsConnected returns whether the Gateway is connected
 func (s Status) IsConnected() bool {
 	switch s {
-	case StatusWaitingForGuilds, StatusReady:
+	case StatusWaitingForHello, StatusIdentifying, StatusWaitingForReady, StatusReady:
```
Skye-31, line 18: It's probably not a good idea to return true for all of these as it may not actually be properly connected. It should only return this for StatusReady. There could be a separate IsSocketConnected() method that tests all of these properties

TopiSenpai: you can simply test for the Status being ready was my idea
as there is only one ready status

TopiSenpai: there is now no usage of it anyway

Skye-31 (approved):

PR #91, opened 2021-09-30
add emoji to role and do some weird pointer shit

caneleex:


discord/role.go
```diff
@@ -39,6 +41,7 @@ type RoleUpdate struct {
 	Color       *int          `json:"color"`
 	Hoist       *bool         `json:"hoist"`
 	Icon        *OptionalIcon `json:"icon,omitempty"`
+	Emoji       *string       `json:"unicode_emoji,omitempty"`
```
caneleex, line 44: actually I don't think this should be omitempty

TopiSenpai (approved):

PR #99, opened 2021-11-01
add support for min/max values

caneleex:


Skye-31 (approved):

TopiSenpai (commented):
not sure if pointers would fit better as you may want to restrict from -15 to 0

TopiSenpai:
instead of pointer we could also use the OptionalInt & OptionalFload64(not there yet)

TopiSenpai (approved):

PR #104, opened 2021-11-10
add support for member time outs

caneleex:
https://github.com/discord/discord-api-docs/pull/4075


json/null_types.go
```diff
@@ -55,3 +57,17 @@ func NewNullFloat() *NullFloat {
 	v := NullFloat(nil)
 	return &v
 }
+
+type NullTime *time.Time
```
TopiSenpai, line 61: I think you miss understood me here
I moved the non discord dependant null types into the json package(float, int, etc...) but `discord.Time`/`discord.NullTime` should still be used

TopiSenpai (approved):
lgtm

PR #105, opened 2021-11-18
add InstallationParams to Application

caneleex:


TopiSenpai (commented):
I know tags are not used atm
But maybe you can add them too just for completeness

discord/application.go
```diff
@@ -6,24 +6,26 @@ import (
...
+	Name                string              `json:"name"`
+	Icon                *string             `json:"icon,omitempty"`
+	Description         string              `json:"description"`
+	RPCOrigins          []string            `json:"rpc_origins"`
+	BotPublic           bool                `json:"bot_public"`
+	BotRequireCodeGrant bool                `json:"bot_require_code_grant"`
+	TermsOfServiceURL   *string             `json:"terms_of_service_url,omitempty"`
+	PrivacyPolicyURL    *string             `json:"privacy_policy_url,omitempty"`
+	InstallationParams  *InstallationParams `json:"install_params"`
+	Tags                *[]string           `json:"tags"`
```
TopiSenpai, line 19:
```suggestion
	Tags                []string            `json:"tags"`
```
this should not be a pointer

caneleex: the `tags` key is missing if there are no tags, it doesn't return an empty array, so it'd have to be nil, wouldn't it?

TopiSenpai: Slices can be nil already so it's fine

TopiSenpai (approved):
lgtm

PR #106, opened 2021-11-18
Webhook refactor

TopiSenpai:
This refractors `discord.Webhook` to be an interface like `discord.ApplicationCommand`/`discord.Component`

Different webhook types have different interactions which this pr implements

This pr also moves the MessageCreate/UpdateBuilder into the discord package as they don't depend on the core package

Same with WebhookMessageCreate/UpdateBuilder
And EmbedBuilder

This also removes the dependency of the webhook package to the core package


TopiSenpai:
Missing: webhook client from `discord.IncomingWebhook`

PR #111, opened 2021-12-02
omit authorization headers where its not needed

TopiSenpai:
send authorization header with bot token only on routes where its needed

### DIFF
--- a/_examples/test/examplebot.go
+++ b/_examples/test/examplebot.go
@@ -26,6 +26,9 @@ var (
 	guildID     = discord.Snowflake(os.Getenv("disgo_guild_id"))
 	adminRoleID = discord.Snowflake(os.Getenv("disgo_admin_role_id"))
 	testRoleID  = discord.Snowflake(os.Getenv("disgo_test_role_id"))
+
+	//go:embed gopher.png
+	gopher []byte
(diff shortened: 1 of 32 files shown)


TopiSenpai:
atm on every request the rest client does the bot token is included
this is bad when you use it for your own requests

there are also some endpoints which does not need authorization where we should not send the header

Skye-31:
these routes are ratelimited significantly more strictly without a bot token, this should be a bot option (disabled by default) rather than done for the user

TopiSenpai:
which routes do you mean by these? 🤔

Skye-31:
invite info by code (/api/vX/invites/:code) for example

TopiSenpai:
then we use those with bot token
can you mark which?

Skye-31:
I don't know of all of them, they are undocumented and subject to change at any point. a client option is the safer way to go here than a user's code randomly breaking

TopiSenpai:
or we just use all endpoints with bot token except interaction & webhook 🥲

Skye-31:
that could work fine

TopiSenpai:
should be better now

Skye-31 (approved):

PR #139, opened 2022-04-09
initial forum channel support

TopiSenpai:
https://github.com/discord/discord-api-docs/pull/4761


discord/channel.go
```diff
@@ -27,8 +27,52 @@ const (
...
 	ChannelTypeGuildStageVoice
 	ChannelTypeGuildDirectory
+	ChannelTypeGuildForum
 )
 
+type ChannelFlags int
+
+const (
+	ChannelFlagsNone   ChannelFlags = 0
+	ChannelFlagsPinned ChannelFlags = 1 << iota
```
Skye-31, line 37: Perhaps rename this to ChannelFlagsThreadPinned to make it clearer?

discord/channel.go
```diff
@@ -909,6 +1042,38 @@ func ApplyGuildIDToChannel(channel GuildChannel, guildID snowflake.Snowflake) Gu
 		c.guildID = guildID
 		return c
 	default:
-		panic("unknown channel type")
+		panic(fmt.Sprintf("invalid channel type: %d", channel.Type()))
```
Skye-31, line 1045: Is panicking the best idea here? It's always possible discord adds a new type and this should not break production bots. It should probably warn instead.

Skye-31 (changes requested):

discord/channel.go
TopiSenpai: it's documented as pinned
maybe they plan to do other stuff with it so I disagree

discord/channel.go
TopiSenpai: hm yeah prob

discord/channel.go
Skye-31: image
This is an example of it, I don't think they do

TopiSenpai: I'd keep it as close to the docs as possible to avoid breaking changes later 
¯\_(ツ)_/¯

Skye-31: The name of an enum is not breaking if discord changes it

TopiSenpai: if they change it so you can pin normal channels it would be

Skye-31: The description of the flag specifically indicates it's for threads
> this thread is pinned to the top of its parent forum channel

TopiSenpai: then we can add an description 
keeping the naming consistent with the docs is still the way to go

TopiSenpai:
this should be tested first before merging

discord/channel.go
Skye-31, line 37: To make it consistent with other flag naming in the library
```suggestion
	ChannelFlagNone   ChannelFlags = 0
	ChannelFlagPinned ChannelFlags = 1 << iota
```

Skye-31 (approved):

TopiSenpai:
there are some new good changes on forum thread creation https://github.com/discord/discord-api-docs/pull/4850

Skye-31 (approved):

PR #142, opened 2022-04-15
fix `withBearerToken` not working

wnxd:


Skye-31:
It's worth noting that the check failure is a result of upstream recent commits, not this PR

TopiSenpai:
I fixed dev feel free to pull in

TopiSenpai (changes requested):
the request header is specifically set before applying the opts to allow overriding the header with `WithHeader` or any other opt

maybe it would be better to make `WithToken` set it directly instead of setting `TokenType` & `Token`?

TopiSenpai:
should be good like this
what do you think @wnxd?

TopiSenpai (approved):

TopiSenpai:
thank you for your contribution

wnxd:
Yeah. Your changes are better.

PR #145, opened 2022-04-20
fix gateway reconnecting bug

TopiSenpai:
it seems like discord can close the websocket with the 1000 code which we treated to not reconnect atm

* reconnect on webscoket close code 1000
* don't reconnect on net.ErrClosed
* move resume logging to debug


TopiSenpai:
this should be ready to go but needs propper testing

PR #150, opened 2022-04-30
transition to uint64 snowflake ids and application command permissions v2

TopiSenpai:
application command permissions v2 have a nice gimmick where you can provide `guild_id - 1` to disallow usage of a command in all channels

unit64 snowflakes were something I was thinking for some time and the transition should be done better now than later


TopiSenpai:
this should be ready but perhaps needs better testing

Skye-31 (approved):

PR #153, opened 2022-05-13
expose sessionid and sequence on gateway, implement both on gatewayimpl

ftqo:
correct my wording if it's wrong xd


TopiSenpai (approved):
Looks fine to me

PR #160, opened 2022-06-03
Fix `getRouteHash` concurrency safety

wnxd:


wnxd:
Since there is no lock for map reading and writing, panic seems to occur under concurrency.

TopiSenpai (changes requested):
I think I'd be happier to not use `sync.Map` with casting
instead use a `sync.Mutex`

wnxd:
> I think I'd be happier to not use `sync.Map` with casting instead use a `sync.Mutex`

Emmmm, I changed it to `sync.Mutex`.

rest/rest_rate_limiter_impl.go
```diff
@@ -96,15 +97,18 @@ func (l *rateLimiterImpl) Reset() {
 	l.bucketsMu = sync.Mutex{}
 	l.global = 0
 	l.hashes = map[*route.APIRoute]routeHash{}
+	l.hashesMu = sync.Mutex{}
```
TopiSenpai, line 100:
```suggestion
```
zero value mutex is already valid

TopiSenpai (changes requested):
besides that small thingy looks good to me

wnxd:
> besides that small thingy looks good to me

I saw that `l.bucketsMu = sync.Mutex{}` was used above. So I added `l.hashes = sync.Mutex{}`

TopiSenpai:
ohh nevermind it should be there yes

I thought this was in the constructor lol

wnxd:
Ok. I add it.

TopiSenpai (approved):

TopiSenpai:
thank you!

PR #170, opened 2022-07-01
Feature/rpc

topi314:


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/disgoorg/disgo
 go 1.18
 
 require (
+	github.com/Microsoft/go-winio v0.5.2
```
YonLiud, line 6: Isn't there a better package for Linux users as well?

TopiSenpai: linux doesn't require an extra package. see https://github.com/disgoorg/disgo/blob/feature/rpc/rpc/pipe_unix.go

dennisrijsdijk:
is there still interest in continuing RPC development? if so, I could look at picking up some work for this.

topi314:
If I recall correctly the status is that most stuff has been implemented, but I wasn't happy with the API design. If you have interest in picking this up I have no objections

topi314:
I am referring to how the send requires a callback to be passed to handle the answer

dennisrijsdijk:
Ah, gotcha. I've seen another library (Rich Presence only) handle it by returning the value, so it should definitely be possible to get working for this as well. I'll have a look into that.

topi314:
that would be the ideal solution, but ideally this would be made 100% type safe using generics

topi314:
if you want to chat more about these changes, feel free to join our discord

Parad1se-py:
what happened to this ;-;

topi314:
no motivation/usecase/other things to do

dennisrijsdijk:
> what happened to this ;-;

lost interest in figuring out Discord's broken docs on all types of messages it could send

Parad1se-py:
honestly, makes sense.

Issue #44, opened 2021-06-28
validation of payloads on rest routes


topi314:
we should validate everything we send to discord & error

example message content too long

topi314:
this is too much work to keep maintained

PR #187, opened 2022-08-15
payload validation

caneleex:


Malix-off:
@topi314 is there any log as of why this PR didn't make it?

topi314:
client side payload validation would be a nice thing to have since this would stop bots making wrong requests

but we ultimately decided against it since its a lot of work to get it initially right and no one had the time for it

Malix-off:
@topi314 Understandable.
Do that mean that #44 should be closed as not planned?

topi314:
maybe, maybe not
its still something I'd like to have so I kept it open

PR #195, opened 2022-08-16
replace DeleteMessageDays with DeleteMessageSeconds and make AddBan take a Duration

caneleex:
https://github.com/discord/discord-api-docs/pull/5219

### DIFF
--- a/discord/ban.go
+++ b/discord/ban.go
@@ -8,5 +8,5 @@ type Ban struct {
 
 // AddBan is used to ban a User (https://discord.com/developers/docs/resources/guild#create-guild-ban-json-params)
 type AddBan struct {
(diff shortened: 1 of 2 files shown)


caneleex:
alternatively, we could change this method to take `discord.AddBan` instead of a value and allow the user to choose which field they want to use before `delete_message_days` gets removed in the next API version.

TopiSenpai:
hm this one is quite annoying as it doesn't break the method signature. What if we instead make the method take `time.Duration`?
sadly this will still not break the signature and could lead to confusion. On the flip side I don't think a lot of people use(if any) so if we break it now it should be fine

TopiSenpai (approved):

PR #196, opened 2022-08-31
improve presence UX

caneleex:
this PR changes presence data to be optional, resulting in not having to pass unwanted data.

before:
```go
bot.WithGatewayConfigOpts(
			gateway.WithIntents(gateway.IntentsNone),
			gateway.WithPresence(gateway.NewListeningPresence("music", discord.OnlineStatusOnline, false)),
)
```
after:
```go
// this yields the same result as the example above
bot.WithGatewayConfigOpts(
			gateway.WithIntents(gateway.IntentsNone),
			gateway.WithPresenceOpts(gateway.WithListeningActivity("music")),
)
```

### DIFF
--- a/gateway/gateway_config.go
+++ b/gateway/gateway_config.go
@@ -184,11 +184,11 @@ func WithRateRateLimiterConfigOpts(opts ...RateLimiterConfigOpt) ConfigOpt {
 // WithPresenceOpts allows to pass initial presence data the bot should display.
 func WithPresenceOpts(opts ...PresenceOpt) ConfigOpt {
 	return func(config *Config) {
-		presence := &MessageDataPresenceUpdate{}
+		presenceUpdate := &MessageDataPresenceUpdate{}
 		for _, opt := range opts {
-			opt(presence)
+			opt(presenceUpdate)
 		}
-		config.Presence = presence
+		config.Presence = presenceUpdate
 	}
 }
 
--- a/gateway/gateway_config.go
+++ b/gateway/gateway_config.go
@@ -60,6 +60,8 @@ func (c *Config) Apply(opts []ConfigOpt) {
 	}
 }
 
+type PresenceOpt func(presenceUpdate *MessageDataPresenceUpdate)
+
 // WithLogger sets the Logger for the Gateway.
 func WithLogger(logger log.Logger) ConfigOpt {
 	return func(config *Config) {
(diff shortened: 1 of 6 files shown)


Skye-31 (approved):

gateway/gateway_config.go
```diff
@@ -60,6 +60,8 @@ func (c *Config) Apply(opts []ConfigOpt) {
 	}
 }
 
+type PresenceOpt func(presenceUpdate *MessageDataPresenceUpdate)
```
TopiSenpai, line 63: one minor thing
Can you move this into `gateway/gateway_messages.go`?

gateway/gateway_messages.go
```diff
@@ -437,57 +437,67 @@ type IdentifyCommandDataProperties struct {
...
-func NewCompetingPresence(name string, status discord.OnlineStatus, afk bool) MessageDataPresenceUpdate {
-	return NewPresence(discord.ActivityTypeCompeting, name, "", status, afk)
+func WithOnlineStatus(status discord.OnlineStatus) PresenceOpt {
+	return func(presence *MessageDataPresenceUpdate) {
+		presence.Status = status
+		if status == discord.OnlineStatusIdle {
+			var since *int64
+			unix := time.Now().Unix()
+			since = &unix
+			presence.Since = since
```
TopiSenpai, line 492:
```suggestion
			since := time.Now().Unix()
			presence.Since = &since
```

TopiSenpai (changes requested):
atm all fields in `MessageDataPresenceUpdate` are not optional. like this all fields get overwritten regardless. 
are they optional(discord docs don't say so)? we need to check else this pr is kinda pointless

gateway/gateway_messages.go
```diff
@@ -437,57 +436,77 @@ type IdentifyCommandDataProperties struct {
...
-			Name: name,
-			Type: activityType,
-		}
-		if activityType == discord.ActivityTypeStreaming && url != "" {
-			activity.URL = &url
-		}
-		activities = append(activities, activity)
-	}
+// WithPlayingActivity creates a new "Playing ..." activity of type discord.ActivityTypeGame
+func WithPlayingActivity(name string) PresenceOpt {
```
TopiSenpai, line 442: should we rename those functions to `WithActivityX` for better discoverability?

caneleex: I think it's safe to assume people these days use intelligent IDEs which will already suggest the methods when using proper keywords such as "activity" or "listening".

TopiSenpai (approved):

PR #199, opened 2022-09-03
replace UpdateSelfNick with UpdateCurrentMember

caneleex:
the [`Modify Current User Nick`](https://discord.com/developers/docs/resources/guild#modify-current-user-nick) endpoint is deprecated.

### DIFF
--- a/discord/member.go
+++ b/discord/member.go
@@ -85,5 +85,5 @@ type MemberUpdate struct {
 
 // CurrentMemberUpdate is used to update the current member
 type CurrentMemberUpdate struct {
-	Nick *string `json:"nick"`
+	Nick string `json:"nick"`
 }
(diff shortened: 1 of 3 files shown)


discord/member.go
```diff
@@ -83,7 +83,7 @@ type MemberUpdate struct {
 	CommunicationDisabledUntil *json.Nullable[time.Time] `json:"communication_disabled_until,omitempty"`
 }
 
-// SelfNickUpdate is used to update your own nick
-type SelfNickUpdate struct {
-	Nick string `json:"nick"`
+// CurrentMemberUpdate is used to update the current member
+type CurrentMemberUpdate struct {
+	Nick *string `json:"nick"`
```
TopiSenpai, line 88: empty string should reset this field

TopiSenpai (changes requested):

discord/member.go
caneleex: passing `nil` is definitely better UX.

TopiSenpai: creating a pointer is simply annoying
I'd prefer to avoid them where possible

TopiSenpai (approved):

PR #204, opened 2022-10-14
add SetFooterIconf, SetImagef, SetThumbnailf and SetURLf

itasli:


TopiSenpai (changes requested):
I think it would be more elegant to just call `SetFooterIcon` from `SetFooterIconf` instead of copying the code

You can also add the last missing ones:
- `SetAuthorNamef`
- `SetAuthorURLf`
- `SetAuthorIconf`
- `SetFooterTextf`

TopiSenpai (approved):

PR #211, opened 2022-11-19
cache refactor

TopiSenpai:


cache/cache_flags.go
```diff
@@ -1,56 +1,56 @@
 package cache
 
-// Flags are used to enable/disable certain internal caches
-type Flags int
+// Types are used to enable/disable certain internal caches
+type Types int
```
caneleex, line 4: I'm not sure I agree with renaming this. why do you think it's suitable to rename it?

TopiSenpai: old one is prob fine agreed

caneleex (approved):

PR #212, opened 2022-11-23
Add pagination limit as a param for GetMembers()

beachasaurus-rex:
Added a pagination limit for `GetMembers()` since discord sets it to 1 by default. [ref](https://discord.com/developers/docs/resources/guild#list-guild-members)

### DIFF
--- a/rest/members.go
+++ b/rest/members.go
@@ -13,7 +13,7 @@ func NewMembers(client Client) Members {
 
 type Members interface {
 	GetMember(guildID snowflake.ID, userID snowflake.ID, opts ...RequestOpt) (*discord.Member, error)
-	GetMembers(guildID snowflake.ID, opts ...RequestOpt) ([]discord.Member, error)
+	GetMembers(guildID snowflake.ID, limit int, opts ...RequestOpt) ([]discord.Member, error)
 	SearchMembers(guildID snowflake.ID, query string, limit int, opts ...RequestOpt) ([]discord.Member, error)
 	AddMember(guildID snowflake.ID, userID snowflake.ID, memberAdd discord.MemberAdd, opts ...RequestOpt) (*discord.Member, error)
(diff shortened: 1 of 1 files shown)


rest/members.go
```diff
@@ -13,7 +13,7 @@ func NewMembers(client Client) Members {
 
 type Members interface {
 	GetMember(guildID snowflake.ID, userID snowflake.ID, opts ...RequestOpt) (*discord.Member, error)
-	GetMembers(guildID snowflake.ID, opts ...RequestOpt) ([]discord.Member, error)
+	GetMembers(guildID snowflake.ID, limit int, opts ...RequestOpt) ([]discord.Member, error)
```
TopiSenpai, line 16: can you also add the `after` param?

rest/members.go
```diff
@@ -40,8 +40,9 @@ func (s *memberImpl) GetMember(guildID snowflake.ID, userID snowflake.ID, opts .
 	return
 }
 
-func (s *memberImpl) GetMembers(guildID snowflake.ID, opts ...RequestOpt) (members []discord.Member, err error) {
-	err = s.client.Do(GetMembers.Compile(nil, guildID), nil, &members, opts...)
+func (s *memberImpl) GetMembers(guildID snowflake.ID, limit int, opts ...RequestOpt) (members []discord.Member, err error) {
+	reqOptions := append(opts, WithQueryParam("limit+", limit))
```
TopiSenpai, line 44: you can pass a map of query params directly to compile. and there is a `+` which isn't needed

TopiSenpai (changes requested):

rest/members.go
beachasaurus-rex: i updated this in the most recent commit

rest/members.go
beachasaurus-rex: i updated this in the most recent commit

TopiSenpai (approved):

PR #221, opened 2022-12-17
feat: webhook URL parse

jolheiser:
Considering Discord doesn't show you the webhook URL and has you copy it as-is, this is a convenience func for parsing a webhook URL wholesale.

As far as I know, they come in the form `https://discord.com/api/webhooks/123456789123456789/foo` (without an API version), at least in from the UI, so this somewhat naively assumes that splitting the URL should yield the ID/token in the 3/4 index respectively.

An alternative would be to instead do a backwards grab of the last two indexes rather than assume 3/4.  
Let me know if you'd prefer that instead.

### DIFF
--- a/webhook/webhook_client_impl.go
+++ b/webhook/webhook_client_impl.go
@@ -12,7 +12,7 @@ import (
 )
 
 // ParseURL parses a webhook URL
-func ParseURL(webhook string, opts ...ConfigOpt) (Client, error) {
+func NewWithURL(webhook string, opts ...ConfigOpt) (Client, error) {
 	u, err := url.Parse(webhook)
 	if err != nil {
 		return nil, err
--- a/webhook/webhook_client_impl.go
+++ b/webhook/webhook_client_impl.go
@@ -2,11 +2,35 @@ package webhook
 
 import (
 	"context"
+	"errors"
+	"net/url"
+	"strings"
 
 	"github.com/disgoorg/disgo/discord"
 	"github.com/disgoorg/disgo/rest"
 	"github.com/disgoorg/snowflake/v2"
 )
+
+// ParseURL parses a webhook URL
+func ParseURL(webhook string, opts ...ConfigOpt) (Client, error) {
+	u, err := url.Parse(webhook)
+	if err != nil {
+		return nil, err
+	}
+
+	parts := strings.FieldsFunc(u.Path, func(r rune) bool { return r == '/' })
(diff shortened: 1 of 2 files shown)


webhook/webhook_client_impl.go
```diff
@@ -2,12 +2,36 @@ package webhook
...
 )
 
+// ParseURL parses a webhook URL
+func ParseURL(webhook string, opts ...ConfigOpt) (Client, error) {
+	u, err := url.Parse(webhook)
+	if err != nil {
+		return nil, err
+	}
+
+	parts := strings.FieldsFunc(u.Path, func(r rune) bool { return r == '/' })
```
TopiSenpai, line 21: why do you not just use `strings.Split`?

jolheiser: `Fields(Func)` will ignore otherwise empty parts, so `/a/b/c/` returns `[a b c]` (3 items) rather than `[ a b c ]` (5 items)
Which makes it easier to reason about than having to deal with ending slashes, etc.

TopiSenpai: ah cool 👍🏻

webhook/webhook_client_impl.go
```diff
@@ -2,12 +2,36 @@ package webhook
...
+	"net/url"
+	"strings"
 
 	"github.com/disgoorg/disgo/discord"
 	"github.com/disgoorg/disgo/rest"
 	"github.com/disgoorg/snowflake/v2"
 )
 
+// ParseURL parses a webhook URL
+func ParseURL(webhook string, opts ...ConfigOpt) (Client, error) {
```
TopiSenpai, line 15: considering you already return the client here this makes more sense I think
```suggestion
func NewWithURL(webhook string, opts ...ConfigOpt) (Client, error) {
```

webhook/webook_test.go
```diff
@@ -0,0 +1,56 @@
...
+			ID:    snowflake.ID(123456789123456789),
+			Token: "foo",
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.URL, func(t *testing.T) {
+			assert := require.New(t)
+
+			c, err := ParseURL(tc.URL)
```
TopiSenpai, line 47:
```suggestion
			c, err := NewWithURL(tc.URL)
```

TopiSenpai (approved):

PR #260, opened 2023-04-04
Pass through logger from config.

iangcarroll:
It appears 2fc89bc438ab907d22fb63f595ebc6ccf50bd02a added a new `Logger` to `StateControllerConfig`, however it was not passed through to the underlying `stateControllerImpl`. It is always `nil` as a result and `NewState`/`UseState` will always panic when called because of this.


TopiSenpai:
thanks for the fix!

PR #266, opened 2023-04-17
Add Reaction() to emojis

caneleex:


discord/emoji.go
```diff
@@ -61,3 +66,14 @@ type PartialEmoji struct {
 	Name     *string       `json:"name"`
 	Animated bool          `json:"animated"`
 }
+
+func (e PartialEmoji) Reaction() string {
+	return reaction(e.ID, *e.Name)
```
TopiSenpai, line 71: We should nil check the name here

TopiSenpai (changes requested):

discord/emoji.go
```diff
@@ -21,6 +22,14 @@ type Emoji struct {
 	Available     bool           `json:"available,omitempty"`
 }
 
+// Reaction returns a string used for manipulating with reactions. May be empty if the length of Name is 0
+func (e Emoji) Reaction() string {
+	if len(e.Name) == 0 {
```
TopiSenpai, line 27:
```suggestion
	if e.Name == ""{
```
Whole `len() == 0` is not wrong
I think checking for an empty string makes more sense

TopiSenpai (approved):

PR #268, opened 2023-04-20
fix(caches): properly calculate permissions' bitfield

ZenShibata:
This calculation logic ported from https://discord.com/developers/docs/topics/permissions#permissions


TopiSenpai:
thank you!

PR #269, opened 2023-05-20
Add shell.nix

RGBCube:


TopiSenpai:
I don't see why this is needed but ty

PR #304, opened 2023-09-01
Update connections

mlnrDev:
new "domain" type + twitter rebranding to X

- [x] https://github.com/discord/discord-api-docs/pull/6395
- [x] https://github.com/discord/discord-api-docs/pull/6394


discord/connection.go
```diff
@@ -31,7 +32,7 @@ const (
 	ConnectionTypeSteam              ConnectionType = "steam"
 	ConnectionTypeTikTok             ConnectionType = "tiktok"
 	ConnectionTypeTwitch             ConnectionType = "twitch"
-	ConnectionTypeTwitter            ConnectionType = "twitter"
+	ConnectionTypeX                  ConnectionType = "twitter"
```
mlnrDev, line 35: @topi314 do we want to break this? I don't think many people use it anyway, but I'm not sure

topi314: uh, we could add an alias to have both, I don't really care tbh

PR #311, opened 2023-09-07
Merge voice and stage channels into one type

mlnrDev:
the structure of both channel types is identical- this greatly reduces duplicate code with almost no difference for the users.

based on https://github.com/disgoorg/disgo/pull/310


topi314 (commented):
I don't think it's a good idea to merge them
this would be a breaking change now & discord might add new stage specific fields in the future

Each channel type has it's own type except `ANNOUNCEMENT_THREAD`, `PUBLIC_THREAD` & `PRIVATE_THREAD` which are the same. So I think we should keep that

PR #324, opened 2023-10-27
Feature/rpc rewrite

dennisrijsdijk:


rpc/client.go
```diff
@@ -26,212 +26,606 @@ type Transport interface {
...
-	Subscribe(event Event, args CmdArgs, handler Handler) error
-	Unsubscribe(event Event, args CmdArgs) error
-	Send(message Message, handler Handler) error
-	Close()
+type responseMessage struct {
+	data MessageData
+	err  error
 }
 
 func NewClient(clientID snowflake.ID, opts ...ConfigOpt) (Client, error) {
```
topi314, line 34:
```suggestion
func New(clientID snowflake.ID, opts ...ConfigOpt) (Client, error) {
```

rpc/client.go
```diff
@@ -26,212 +26,606 @@ type Transport interface {
...
-		return nil, ctx.Err()
-	case <-client.readyChan:
+		return ctx.Err()
+	case <-c.readyChan:
 	}
+	return nil
+}
 
-	return client, nil
+func (c *Client) sendBytes(r io.Reader) error {
```
topi314, line 85: this has only one usage. why is this here?

rpc/client.go
```diff
@@ -26,212 +26,606 @@ type Transport interface {
...
+	}
 
-	eventHandlers   map[Event]Handler
-	commandHandlers map[string]internalHandler
+	if res, err := c.send(Message{
+		Cmd:  CmdAuthorize,
+		Args: args,
+	}); err != nil {
+		return "", err
+	} else {
```
topi314, line 119: `else` = smelly

rpc/client.go
```diff
@@ -26,212 +26,606 @@ type Transport interface {
...
+		return res.(CmdRsAuthorize).Code, nil
```
topi314, line 120:
```go
		return res.(CmdRsAuthorize).Code, nil
```
is there any chance `res` will not be of type `CmdRsAuthorize`?

rpc/client.go
```diff
@@ -26,212 +26,606 @@ type Transport interface {
...
+	}); err != nil {
+		return nil, err
+	} else {
+		return res.(CmdRsGetGuilds).Guilds, nil
+	}
 }
 
-func (c *clientImpl) User() discord.User {
-	return c.user
+func (c *Client) GetChannel(channelID snowflake.ID) (CmdRsGetChannel, error) {
```
topi314, line 165: All these methods should return `(*T, error)`

rpc/cmd.go
```diff
@@ -223,16 +235,21 @@ type ShortcutKeyCombo struct {
...
+	NoiseSuppression     *bool              `json:"noise_suppression,omitempty"`
+	QOS                  *bool              `json:"qos,omitempty"`
+	SilenceWarning       *bool              `json:"silence_warning,omitempty"`
+	Deaf                 *bool              `json:"deaf,omitempty"`
+	Mute                 *bool              `json:"mute,omitempty"`
+}
+
+func (EmptyArgs) cmdArgs() {}
+
+type EmptyArgs struct {
```
topi314, line 252: can't you use nil instead?

rpc/cmd.go
```diff
@@ -131,14 +142,15 @@ func (CmdRsSetUserVoiceSettings) messageData() {}
...
+	Timeout   int          `json:"timeout,omitempty"`
 	Force     bool         `json:"force"`
+	Navigate  bool         `json:"navigate"`
 }
 
 func (CmdArgsSelectVoiceChannel) cmdArgs() {}
 
 type CmdRsSelectVoiceChannel struct {
-	PartialChannel
+	*PartialChannel
```
topi314, line 153: is this a pointer because the result can be null?

rpc/embed.go
```diff
@@ -0,0 +1,37 @@
...
+	Thumbnail      *EmbedResource         `json:"thumbnail,omitempty"`
+	Video          *EmbedResource         `json:"video,omitempty"`
+	Provider       *discord.EmbedProvider `json:"provider,omitempty"`
+	Author         *discord.EmbedAuthor   `json:"author,omitempty"`
+}
+
+type Attachment struct {
+	*discord.Attachment
+	Spoiler bool `json:"spoiler,omitempty"`
+}
```
topi314, line 37: is embed so different from `discord.Embed`?

rpc/message.go
```diff
@@ -268,29 +269,147 @@ type CmdArgs interface {
 	cmdArgs()
 }
 
-type Event string
+type Event struct {
+	EventType
+	CmdArgs
+}
+
+func BuildIncomingEvent(eventType EventType, data MessageData) (Event, error) {
```
topi314, line 277:
```suggestion
func UnmarshalEvent(eventType EventType, data MessageData) (Event, error) {
```

topi314 (changes requested):
in general this looks nice

I'm unsure if I like how subscribing to events work.
Maybe it should work more like in the normal bot where you register handlers via `AddEventListeners` and the `subscribe` method won't take a handler

I'm open for discussion here

rpc/client.go
dennisrijsdijk: in old rpc branch this was the same (`(c *clientImpl) send(r io.Reader) error`) with only one usage. This has been inlined in [f531263](https://github.com/dennisrijsdijk/disgo/commit/f531263bd0a55277578b3e5055aef04f5bcab81d)

rpc/client.go
dennisrijsdijk: unless there's an error (checked before accessing `res`) it's always `CmdRsAuthorize`

rpc/cmd.go
dennisrijsdijk: Correct. When leaving a voice channel, the result is null.

topi314: then you wont need this to be a pointer
```suggestion
	PartialChannel
```

rpc/client.go
dennisrijsdijk: fixed in [4761252](https://github.com/dennisrijsdijk/disgo/commit/47612524627647d77f406eb356b1f85dfa952f1e)

rpc/message.go
dennisrijsdijk: fixed in [4761252](https://github.com/dennisrijsdijk/disgo/commit/47612524627647d77f406eb356b1f85dfa952f1e)

rpc/cmd.go
dennisrijsdijk: for `CmdRsSelectVoiceChannel` that makes sense. What about `CmdRsGetSelectedVoiceChannel`? Should that remain a pointer (`if channel == nil` to detect no channel) or should I also remove the pointer there?  (`if channel.ID == 0` to detect no channel)

topi314: that shouldnt be a pointer either
the struct will be nil in this case

dennisrijsdijk: structs can't be `nil` though. it defaults to a struct with all 0 values

topi314: ah right we don't use a pointer here https://github.com/dennisrijsdijk/disgo/blob/feature/rpc-rewrite/rpc/message.go#L188
then this is fine and we can check for `id == 0`

dennisrijsdijk: Then again, I'm working on the earlier comment that everything should return `(*T, error)` so I could add that logic into the function there as well.

topi314: sounds good to me

rpc/client.go
dennisrijsdijk: Did you want this on Authorize as well?
eg. from `Authorize(scopes []discord.OAuth2Scope, rpcToken string, username string) (string, error)`
to `Authorize(scopes []discord.OAuth2Scope, rpcToken string, username string) (*string, error)`
or `Authorize(scopes []discord.OAuth2Scope, rpcToken string, username string) (*CmdRsAuthorize, error)`

topi314: yes

dennisrijsdijk: Resolved in [4ef8ffc](https://github.com/dennisrijsdijk/disgo/commit/4ef8ffc03ba58f70cb22951e155ddf66e18ef4e6)

rpc/cmd.go
dennisrijsdijk: Resolved in [4ef8ffc](https://github.com/dennisrijsdijk/disgo/commit/4ef8ffc03ba58f70cb22951e155ddf66e18ef4e6)

rpc/client.go
dennisrijsdijk: Resolved in [4ef8ffc](https://github.com/dennisrijsdijk/disgo/commit/4ef8ffc03ba58f70cb22951e155ddf66e18ef4e6)

rpc/cmd.go
dennisrijsdijk: Resolved in [ca68a83](https://github.com/dennisrijsdijk/disgo/commit/ca68a83591c1ade9626007093a8638ccebbde717)

rpc/embed.go
dennisrijsdijk: changed in [fb13deb](https://github.com/dennisrijsdijk/disgo/commit/fb13deb455f73a7061e86ba02f7294bd2bd87020), looking for more feedback.

rpc/message.go
```diff
@@ -268,29 +339,147 @@ type CmdArgs interface {
 	cmdArgs()
 }
 
-type Event string
+type Event struct {
```
topi314, line 342: I'm not sure what the purpose of this is

topi314 (changes requested):
I think we need to talk about the event subscriptions

rpc/message.go
```diff
@@ -1,26 +1,97 @@
...
+	m.Attachments = v.Attachments
+	m.Blocked = v.Blocked
+	m.Bot = v.Bot
+	m.Embeds = v.Embeds
+	m.Mentions = v.Mentions
+	m.Nick = v.Nick
+	return nil
+}
+
+type message struct {
```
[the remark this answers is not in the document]
dennisrijsdijk: Would you suggest I rename this or Message (as in ChannelMessage)?

rpc/message.go
dennisrijsdijk: Related atm to keeping track of subscribed events (event name + args) and handlers. to be revisited when we decide on the approach for events going forward.

PR #334, opened 2024-03-01
enhance: readibility

Malix-off:


Malix-off:
Ah ye right.
Do you plan on making the upgrade to 1.22?
That loop is the only one I found that can be upgraded to a ranging over integer loop (appart from https://github.com/disgoorg/disgo/blob/842ff0cd02829815f411d44d9c66f865a94573be/_examples/webhook/example.go#L34 which starts at 1)

topi314:
I plan to eventually update the go version when 1.21 hits EOL (should be with go 1.23 release) or there are compelling new features

PR #336, opened 2024-03-08
Disgoorg master

snekROmonoro:


topi314:
you know, if there is an issue you can always make a pr with a fix & I'm gonna happily merge it without you needing to hardfork 💀

snekROmonoro:
> you know, if there is an issue you can always make a pr with a fix & I'm gonna happily merge it without you needing to hardfork 💀

yeah ik dw LOL, didn't mean to do this pr, it was done by accident

topi314:
then why are you hard forking and not opening a simple pr with a fix

snekROmonoro:
ig i wanted to hard fork it, idek why i had done it in the past

topi314:
* making those fields `omitempty` is prob completely valid and just an oversight on our side
* adding the actual message data to the error message is valid too and we actually wanted to make that change but just forgot about it
* as for the snowflake changes you want the `"` quotes to be optional so you can also parse number ids from json?

snekROmonoro:
> * as for the snowflake changes you want the `"` quotes to be optional so you can also parse number ids from json?
i remember printing out errors where it failed to parse (valid) snowflakes due to `"` missing

topi314:
then I recommend you open an issue or let us know on our discord so we can look into it instead of doing a yolo

snekROmonoro:
will do in the future 👍🏻

PR #343, opened 2024-03-19
Add TeamRole

mlnrDev:
we seem to have missed this somehow. also, imo, we should get rid of the `Permissions` field since it was removed and never documented properly in the first place: https://github.com/discord/discord-api-docs/pull/6372#discussion_r1304483584

- [x] https://github.com/discord/discord-api-docs/pull/6372
- [x] https://github.com/discord/discord-api-docs/pull/6384

### DIFF
--- a/discord/application.go
+++ b/discord/application.go
@@ -241,11 +241,10 @@ func (t Team) CreatedAt() time.Time {
 }
 
 type TeamMember struct {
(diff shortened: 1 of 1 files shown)


topi314 (approved):
then also remove it in this pr thanks

PR #345, opened 2024-03-20
Refactor oauth params into a struct 

mlnrDev:
with the growing amount of oauth parameters, it only makes sense to group them together as much as possible without breaking changes in the future.

based on and blocked by https://github.com/disgoorg/disgo/pull/341


oauth2/client_impl.go
```diff
@@ -47,30 +47,26 @@ func (c *clientImpl) StateController() StateController {
...
-	if disableGuildSelect {
-		values["disable_guild_select"] = true
+		"client_id":            c.id,
+		"redirect_uri":         params.RedirectURI,
+		"response_type":        "code",
+		"scope":                discord.JoinScopes(params.Scopes),
+		"state":                state,
+		"permissions":          params.Permissions,
+		"disable_guild_select": params.DisableGuildSelect,
+		"integration_type":     params.IntegrationType,
```
topi314, line 65: optional fields should be excluded in the url when they are not set like previously

topi314 (changes requested):

topi314 (approved):

PR #350, opened 2024-04-24
Add message forwarding

sebm253:
- [x] https://github.com/discord/discord-api-docs/pull/6818
- [x] clarification on `message_snapshot` fields and partiality
- [x] https://github.com/discord/discord-api-docs/pull/6833

### DIFF
--- a/discord/message.go
+++ b/discord/message.go
@@ -102,6 +102,7 @@ type Message struct {
 	Type                 MessageType           `json:"type"`
 	Flags                MessageFlags          `json:"flags"`
 	MessageReference     *MessageReference     `json:"message_reference,omitempty"`
+	MessageSnapshot      *Message              `json:"message_snapshot,omitempty"`
 	Interaction          *MessageInteraction   `json:"interaction,omitempty"`
 	WebhookID            *snowflake.ID         `json:"webhook_id,omitempty"`
 	Activity             *MessageActivity      `json:"activity,omitempty"`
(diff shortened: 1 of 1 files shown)


discord/message.go
```diff
@@ -102,6 +102,7 @@ type Message struct {
 	Type                 MessageType           `json:"type"`
 	Flags                MessageFlags          `json:"flags"`
 	MessageReference     *MessageReference     `json:"message_reference,omitempty"`
+	MessageSnapshot      *Message              `json:"message_snapshot,omitempty"`
```
lytefast, line 105: Not sure if you want to map onto `Message` directly here since the fields returned here are a subset. May or may not work out for this project depending on strictness (e.g. `author` is not returned, but is a required field on Message in general)

lytefast (commented):
LGTM, but I'm not familiar with this code base.

discord/message.go
mlnrDev: > Not sure if you want to map onto `Message` directly here since the fields returned here are a subset. May or may not work out for this project depending on strictness (e.g. `author` is not returned, but is a required field on Message in general)

that's why I mentioned clarification on the fields present in the PR description- it'd be helpful to know to perhaps make a custom type for fields that are guaranteed to be in the snapshots, if there are only a handful

Coolmandre:
So what this means I. English

PR #366, opened 2024-06-24
enhance: refactor disgo.OS

Malix-off:


topi314:
What is the reason for this change?

Malix-off:
`runtime.GOOS` already output the desired output from `disgo.getOS()` while making it even more portable

See https://github.com/disgoorg/disgo/pull/366#issue-2370847615#references

topi314:
while I can't exactly tell you why this was introduced in the first place since it's been over 2y lol, it was prob to make it easily overwritable, the structure has been quite different back then and this seems to has stuck with us since.

These days you can overwrite this differently so it's prob fine to get completely rid off the const & replace it with `runtime.GOOS` in https://github.com/disgoorg/disgo/blob/master/disgo.go#L108

Malix-off:
Done

topi314 (approved):

topi314:
thanks

PR #367, opened 2024-06-24
enhance: `disgo` consts - composed

Malix-off:
This is me being very autistic


topi314 (commented):
I don't agree with this change since the go module, git url & library name are not necessarily connected together. go modules can be served via a custom domain as an example which I might intend in the future.

Keeping them as is makes the most sense imo

Malix-off:
Understandable

PR #377, opened 2024-08-12
Make the Audit Log Change Key typed and add constants for its possible values

myrkvi:
Changes `AuditLogChange.Key` to be `AuditLogChangeKey` instead of `string`, and adds constants for its valid values.

This should make it a bit easier to match desired keys, since they will be defined in the library.


topi314 (approved):

topi314:
thanks good timing on this too

PR #379, opened 2024-08-13
Change GuildPrivateThreadCreate to have NonInvitable field instead of Invitable

myrkvi:
The current implementation has the field `Invitable` and the JSON annotation `omitempty`, which results in the field being omitted when it is false. Not explicitly setting an invitable state should result in it being invitable, and as such, having `NonInvitable` achieves this.


topi314:
I'd rather have this be a pointer instead to be consistent with other fields & the discord docs

myrkvi:
> I'd rather have this be a pointer instead to be consistent with other fields & the discord docs

Definitely a saner way to do it :sweat_smile:

topi314:
sane is arguable in this case, getting pointers from primitives is annoying after all lol

topi314:
thanks

PR #381, opened 2024-08-16
Implement "List Active Guild Threads"

myrkvi:
Creates a method that implements the [`/guilds/{guild.id}/threads/active`][ref] endpoint of the Discord API.

[ref]: https://discord.com/developers/docs/resources/guild#list-active-guild-threads

### DIFF
--- a/rest/rest_endpoints.go
+++ b/rest/rest_endpoints.go
@@ -192,6 +192,7 @@ var (
 	GetPublicArchivedThreads        = NewEndpoint(http.MethodGet, "/channels/{channel.id}/threads/archived/public")
 	GetPrivateArchivedThreads       = NewEndpoint(http.MethodGet, "/channels/{channel.id}/threads/archived/private")
 	GetJoinedPrivateArchivedThreads = NewEndpoint(http.MethodGet, "/channels/{channel.id}/users/@me/threads/archived/private")
+	GetActiveGuildThreads           = NewEndpoint(http.MethodGet, "/guilds/{guild.id}/threads/active")
 )
 
 // Messages
--- a/discord/guild.go
+++ b/discord/guild.go
@@ -384,3 +384,8 @@ type GuildPrune struct {
 type GuildPruneResult struct {
 	Pruned *int `json:"pruned"`
 }
+
+type GuildActiveThreads struct {
+	Threads []GuildThread  `json:"threads"`
+	Members []ThreadMember `json:"members"`
+}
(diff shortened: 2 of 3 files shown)


rest/rest_endpoints.go
```diff
@@ -60,6 +60,7 @@ var (
 	CreateGuildChannel     = NewEndpoint(http.MethodPost, "/guilds/{guild.id}/channels")
 	GetGuildChannels       = NewEndpoint(http.MethodGet, "/guilds/{guild.id}/channels")
 	UpdateChannelPositions = NewEndpoint(http.MethodPatch, "/guilds/{guild.id}/channels")
+	GetActiveGuildThreads  = NewEndpoint(http.MethodGet, "/guilds/{guild.id}/threads/active")
```
sebm253, line 63: can you please move this into the part with Thread endpoints? L181

rest/guilds.go
```diff
@@ -123,6 +124,11 @@ func (s *guildImpl) UpdateChannelPositions(guildID snowflake.ID, guildChannelPos
 	return s.client.Do(UpdateChannelPositions.Compile(nil, guildID), guildChannelPositionUpdates, nil, opts...)
 }
 
+func (s *guildImpl) GetActiveGuildThreads(guildID snowflake.ID, opts ...RequestOpt) (activeThreads discord.GuildActiveThreads, err error) {
```
sebm253, line 127: this should return a pointer of the struct

rest/guilds.go
```diff
@@ -27,6 +27,7 @@ type Guilds interface {
 	CreateGuildChannel(guildID snowflake.ID, guildChannelCreate discord.GuildChannelCreate, opts ...RequestOpt) (discord.GuildChannel, error)
 	GetGuildChannels(guildID snowflake.ID, opts ...RequestOpt) ([]discord.GuildChannel, error)
 	UpdateChannelPositions(guildID snowflake.ID, guildChannelPositionUpdates []discord.GuildChannelPositionUpdate, opts ...RequestOpt) error
+	GetActiveGuildThreads(guildID snowflake.ID, opts ...RequestOpt) (discord.GuildActiveThreads, error)
```
sebm253, line 30: same here - pointer

rest/rest_endpoints.go
myrkvi: Sure! Just to confirm, this is the case even though it is part of the Guilds endpoints on Discord's end?

sebm253: correct - we separate them like this even though they share a common endpoint

rest/guilds.go
```diff
@@ -27,6 +27,7 @@ type Guilds interface {
 	CreateGuildChannel(guildID snowflake.ID, guildChannelCreate discord.GuildChannelCreate, opts ...RequestOpt) (discord.GuildChannel, error)
 	GetGuildChannels(guildID snowflake.ID, opts ...RequestOpt) ([]discord.GuildChannel, error)
 	UpdateChannelPositions(guildID snowflake.ID, guildChannelPositionUpdates []discord.GuildChannelPositionUpdate, opts ...RequestOpt) error
+	GetActiveGuildThreads(guildID snowflake.ID, opts ...RequestOpt) (*discord.GuildActiveThreads, error)
```
sebm253, line 30: it would make sense to also move this func into rest/threads

sebm253 (changes requested):

sebm253 (approved):
thanks :)

PR #390, opened 2024-08-26
Add entry point commands + callback updates

sebm253:
- [x] https://github.com/discord/discord-api-docs/pull/7106


rest/interactions.go
```diff
@@ -33,13 +34,23 @@ func (s *interactionImpl) GetInteractionResponse(interactionID snowflake.ID, int
...
 
-func (s *interactionImpl) CreateInteractionResponse(interactionID snowflake.ID, interactionToken string, interactionResponse discord.InteractionResponse, opts ...RequestOpt) error {
+// CreateInteractionResponse calls CreateInteractionResponseWithCallback with `withCallback: false`.
+// The signature of this function might be adjusted in the future.
+func (s *interactionImpl) CreateInteractionResponse(interactionID snowflake.ID, interactionToken string, interactionResponse discord.InteractionResponse, opts ...RequestOpt) (err error) {
+	_, err = s.CreateInteractionResponseWithCallback(interactionID, interactionToken, interactionResponse, false, opts...)
+	return
+}
+
+func (s *interactionImpl) CreateInteractionResponseWithCallback(interactionID snowflake.ID, interactionToken string, interactionResponse discord.InteractionResponse, withCallback bool, opts ...RequestOpt) (callback *discord.InteractionCallbackResponse, err error) {
```
sebm253, line 44: I added a separate func so we don't break the current signature

discord/application_command.go
```diff
@@ -435,3 +444,131 @@ func (c MessageCommand) CreatedAt() time.Time {
...
+	name                     string
+	nameLocalizations        map[Locale]string
+	nameLocalized            string
+	defaultMemberPermissions Permissions
+	dmPermission             bool
+	nsfw                     bool
+	integrationTypes         []ApplicationIntegrationType
+	contexts                 []InteractionContextType
+	version                  snowflake.ID
+	handler                  *EntryPointCommandHandlerType
```
sebm253, line 463: if you create an entry point command without providing the `handler` field, subsequent fetches of the command will not have the field (I have no clue)

rest/interactions.go
```diff
@@ -42,6 +45,18 @@ func (s *interactionImpl) CreateInteractionResponse(interactionID snowflake.ID,
 	return s.client.Do(CreateInteractionResponse.Compile(nil, interactionID, interactionToken), body, nil, opts...)
 }
 
+func (s *interactionImpl) CreateInteractionResponseWithCallback(interactionID snowflake.ID, interactionToken string, interactionResponse discord.InteractionResponse, opts ...RequestOpt) (callback *discord.InteractionCallbackResponse, err error) {
```
sebm253, line 48: I added a separate func so we don't break the current signature

discord/application_command.go
topi314: pretty sure this shouldn't matter then?
since the enum is 1 & 2 0 will indicate that it's missing and we don't need a pointer

discord/application_command_create.go
```diff
@@ -107,3 +107,36 @@ func (c MessageCommandCreate) CommandName() string {
...
+type EntryPointCommandCreate struct {
+	Name                     string                      `json:"name"`
+	NameLocalizations        map[Locale]string           `json:"name_localizations,omitempty"`
+	DefaultMemberPermissions *json.Nullable[Permissions] `json:"default_member_permissions,omitempty"`
+	// Deprecated: Use Contexts instead
+	DMPermission     *bool                         `json:"dm_permission,omitempty"`
+	IntegrationTypes []ApplicationIntegrationType  `json:"integration_types,omitempty"`
+	Contexts         []InteractionContextType      `json:"contexts,omitempty"`
+	NSFW             *bool                         `json:"nsfw,omitempty"`
+	Handler          *EntryPointCommandHandlerType `json:"handler,omitempty"`
```
topi314, line 120: same here, this does not need to be a pointer I think, `omitempty` should omit the 0

discord/application_command.go
```diff
@@ -435,3 +444,131 @@ func (c MessageCommand) CreatedAt() time.Time {
...
+
+func (c EntryPointCommand) Contexts() []InteractionContextType {
+	return c.contexts
+}
+
+func (c EntryPointCommand) Version() snowflake.ID {
+	return c.version
+}
+
+func (c EntryPointCommand) Handler() *EntryPointCommandHandlerType {
```
topi314, line 559: does this need to be a method? we usually have fields which aren't part of the interface as public fields

topi314 (approved):

PR #402, opened 2024-11-30
Add GetEntitlement and new GetEntitlements param

sebm253:
- [x] https://github.com/discord/discord-api-docs/pull/7279

TODO:
- [x] move params into a struct

### DIFF
--- a/rest/query_params.go
+++ b/rest/query_params.go
@@ -2,48 +2,10 @@ package rest
 
 import (
 	"github.com/disgoorg/disgo/discord"
-	"github.com/disgoorg/disgo/internal/slicehelper"
-	"github.com/disgoorg/snowflake/v2"
 )
 
 // QueryParams serves as a generic interface for implementations of rest endpoint query parameters.
 type QueryParams interface {
 	// ToQueryValues transforms fields from the QueryParams interface implementations into discord.QueryValues.
 	ToQueryValues() discord.QueryValues
-}
-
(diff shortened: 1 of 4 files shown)


rest/query_params.go
```diff
@@ -2,10 +2,48 @@ package rest
...
+		queryValues["after"] = p.After
+	}
+	if p.Limit != 0 {
+		queryValues["limit"] = p.Limit
+	}
+	if p.GuildID != 0 {
+		queryValues["guild_id"] = p.GuildID
+	}
+	return queryValues
+}
```
topi314, line 49: shouldn't we move this to where the endpoint is declared?
we can have common ones in here tho

topi314 (changes requested):

rest/query_params.go
sebm253: thought about it, didn't really like that since it would clutter all endpoint declarations with partly irrelevant code, it would also help with seeing them all in one place

topi314: better than having it sit somewhere random place
that place would get cluttered fast

sebm253: it would be meant to get cluttered :-) don't think it's random considering the name of the file is "query_params"

topi314: it would be kinda random since I would expect things which are connected to be at the same place

sebm253: yeah I get that, can't really decide since it makes sense but at the same time I don't want to put it in between endpoint declarations

topi314: just put it above or below the interface

topi314 (approved):

PR #404, opened 2024-12-01
Introduce ItemsResponse

sebm253:


rest/items_response.go
```diff
@@ -0,0 +1,17 @@
...
+
+import "github.com/disgoorg/json"
+
+// ItemsResponse serves as a utility to flatten objects wrapped in an "items" object into a slice of objects.
+type ItemsResponse[T any] []T
+
+func (rs *ItemsResponse[T]) UnmarshalJSON(data []byte) error {
+	var response struct {
+		Items []T `json:"items"`
+	}
```
sebm253, line 11: would moving this into its own type make more sense?

rest/soundboard_sounds.go
```diff
@@ -13,7 +13,7 @@ func NewSoundboardSounds(client Client) SoundboardSounds {
 
 type SoundboardSounds interface {
 	GetSoundboardDefaultSounds(opts ...RequestOpt) ([]discord.SoundboardSound, error)
-	GetGuildSoundboardSounds(guildID snowflake.ID, opts ...RequestOpt) ([]discord.SoundboardSound, error)
+	GetGuildSoundboardSounds(guildID snowflake.ID, opts ...RequestOpt) (ItemsResponse[discord.SoundboardSound], error)
```
topi314, line 16: why break this?

rest/applications.go
```diff
@@ -42,7 +42,7 @@ type Applications interface {
 	DeleteTestEntitlement(applicationID snowflake.ID, entitlementID snowflake.ID, opts ...RequestOpt) error
 	ConsumeEntitlement(applicationID snowflake.ID, entitlementID snowflake.ID, opts ...RequestOpt) error
 
-	GetApplicationEmojis(applicationID snowflake.ID, opts ...RequestOpt) ([]discord.Emoji, error)
+	GetApplicationEmojis(applicationID snowflake.ID, opts ...RequestOpt) (ItemsResponse[discord.Emoji], error)
```
topi314, line 45: why break this?

topi314 (changes requested):
I don't think `ItemResponse` should be exposed to the outside,
this type should be unexported imo

rest/soundboard_sounds.go
sebm253: doesn't break anything and we need the unmarshaller

rest/applications.go
sebm253: https://github.com/disgoorg/disgo/pull/404#discussion_r1865044552

rest/soundboard_sounds.go
topi314: you changed the return type, this is breaking

sebm253: it's still a slice

topi314: just unexport the type and make it an internal matter, no need to expose this outside

sebm253:
closing this for now

PR #409, opened 2025-01-18
Add Acknowledge function to ApplicationCommandInteractionCreate for HTTP server interactions

BowsiePup:
This pull request introduces a new function to  `ApplicationCommandInteractionCreate`

`ApplicationCommandInteractionCreate.Acknowledge()`

#### New Interaction Response Type:
`InteractionResponseTypeAcknowledge`. A Strictly internal iota to identify this response type.

#### New function:
`ApplicationCommandInteractionCreate.Acknowledge()`. This method is used to respond to the original HTTP request from discord with StatusAccepted (202). 
This is not like `ApplicationCommandInteractionCreate.DeferCreateMessage()`. You are still required to respond with a new HTTP request within 3 seconds to respond to the interaction.
Unlike `ApplicationCommandInteractionCreate.CreateMessage()`, you are able to access the response which will contain error or the sent message when you send `event.Client().Rest().CreateInteractionResponse()`

Source to this can be found [here](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-callback)
`If you send this request for an interaction received over HTTP, respond to the original HTTP request with a 202 and no body.`

This is the preferred way to respond to Interactions as you can receive the message created in the response if u need it and errors.

### DIFF
--- a/discord/interaction_response.go
+++ b/discord/interaction_response.go
@@ -5,7 +5,9 @@ type InteractionResponseType int
 
 // Constants for the InteractionResponseType(s)
 const (
-	InteractionResponseTypePong InteractionResponseType = iota + 1
+	// This is stricly internal and will never be sent to discord. It is used to indicate that the HTTP response should be 202 Accepted
+	InteractionResponseTypeAcknowledge InteractionResponseType = iota
+	InteractionResponseTypePong        InteractionResponseType = iota
 	_
 	_
 	InteractionResponseTypeCreateMessage
--- a/discord/interaction_response.go
+++ b/discord/interaction_response.go
@@ -5,9 +5,12 @@ type InteractionResponseType int
 
 // Constants for the InteractionResponseType(s)
 const (
-	// This is stricly internal and will never be sent to discord. It is used to indicate that the HTTP response should be 202 Accepted
-	InteractionResponseTypeAcknowledge InteractionResponseType = iota
-	InteractionResponseTypePong        InteractionResponseType = iota
+	// InteractionResponseTypeAcknowledge is stricly internal and will never be sent to discord.
+	//
+	// It is used to indicate that the HTTP response should be 202 Accepted
+	InteractionResponseTypeAcknowledge InteractionResponseType = -1
+
+	InteractionResponseTypePong InteractionResponseType = iota + 1
 	_
 	_
 	InteractionResponseTypeCreateMessage
--- a/discord/interaction_response.go
+++ b/discord/interaction_response.go
@@ -3,13 +3,13 @@ package discord
 // InteractionResponseType indicates the type of slash command response, whether it's responding immediately or deferring to edit your response later
 type InteractionResponseType int
 
+// InteractionResponseTypeAcknowledge is stricly internal and will never be sent to discord.
+//
+// It is used to indicate that the HTTP response should be 202 Accepted
+const InteractionResponseTypeAcknowledge InteractionResponseType = -1
+
 // Constants for the InteractionResponseType(s)
 const (
-	// InteractionResponseTypeAcknowledge is stricly internal and will never be sent to discord.
-	//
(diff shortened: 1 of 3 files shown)


discord/interaction_response.go
```diff
@@ -5,7 +5,9 @@ type InteractionResponseType int
 
 // Constants for the InteractionResponseType(s)
 const (
-	InteractionResponseTypePong InteractionResponseType = iota + 1
+	// This is stricly internal and will never be sent to discord. It is used to indicate that the HTTP response should be 202 Accepted
+	InteractionResponseTypeAcknowledge InteractionResponseType = iota
+	InteractionResponseTypePong        InteractionResponseType = iota
```
topi314, line 10: `iota` is only needed for the first entry in a `const` block
that being said I think i would prefer a magic value of `-1` instead of `0`.

Your comment should also be prefixed with `InteractionResponseTypeAcknowledge`

events/interaction_events.go
```diff
@@ -42,6 +42,20 @@ func (e *ApplicationCommandInteractionCreate) Guild() (discord.Guild, bool) {
 	return discord.Guild{}, false
 }
 
+// Acknowledge acknowledges the interaction.
+//
+// This is used strictly for acknowledging the HTTP interaction request from discord. This responds with 202 Accepted.
+//
+// This does not produce a visible loading state to the user.
+// You are expected to send a new http request within 3 seconds to respond to the interaction.
```
topi314, line 50: can you link here to the methods to create this http request?

httpserver/server.go
```diff
@@ -167,6 +167,15 @@ func HandleInteraction(publicKey PublicKey, logger *slog.Logger, handleFunc Even
 		defer cancel()
 		select {
 		case response := <-responseChannel:
+
+			// if we only acknowledge the interaction, we don't need to send a response body
+			// we just need to send a 202 Accepted status
+			if response.Type == discord.InteractionResponseTypeAcknowledge {
+				w.WriteHeader(http.StatusAccepted)
+				w.Write(nil)
```
topi314, line 175: I don't think this write is needed

topi314 (changes requested):
Looks like a good addition. Can you also add the `Acknowledge` method to the other interaction events?

BowsiePup:
All requested changes have been added.

Autocomplete does not as it does not have a Interaction Callback Type and you are expected to respond via the http response on the original request.

events/interaction_events.go
```diff
@@ -42,6 +42,23 @@ func (e *ApplicationCommandInteractionCreate) Guild() (discord.Guild, bool) {
 	return discord.Guild{}, false
 }
 
+// Acknowledge acknowledges the interaction.
+//
+// This is used strictly for acknowledging the HTTP interaction request from discord. This responds with 202 Accepted.
+//
+// When using this, your first http request must be <event>.Client().Rest().CreateInteractionResponse()
```
topi314, line 49: you can create a go doc link via `[rest.Interactions.CreateInteractionResponse]` afaik
this should also include `[rest.Interactions.CreateInteractionResponseWithCallback]`

topi314 (changes requested):

events/interaction_events.go
topi314: I am pretty sure you don't even need to link to pkg.go.dev

BowsiePup:
A few issues fixed:

Auto complete does have an interaction type: `APPLICATION_COMMAND_AUTOCOMPLETE_RESULT 8`
Added the function to `AutocompleteInteractionCreate`

Moved `InteractionResponseTypeAcknowledge` out of the iota const as it broke the iota making all items be one extra then they should be. I moved it out to keep the `iota + 1` so it still matches with discord docs.

topi314 (approved):

topi314:
thanks

PR #413, opened 2025-02-04
fix: Pass activity by reference in `ActivityOpt`

apricotbucket28:
Without the reference, assignments to the struct are unused.


topi314:
good catch ty

topi314:
this will be included in the next release, feel free to use master until then

PR #414, opened 2025-02-04
fix: Use `OpenShardManager` in proxy example

apricotbucket28:
Since the bot is configured with a shard manager (`bot.WithShardManagerConfigOpts`), `OpenShardManager` must be used instead of `OpenGateway`.


topi314 (approved):

topi314:
thanks!

PR #417, opened 2025-02-08
Fix incorrect initialization of shard IDs in `BuildClient`

apricotbucket28:
Fixes a off-by-one error when initializing shardIDs the `BuildClient`.

Currently, if `gatewayBotRs.Shards` is 5, then shardIDs first gets initialized to:
```go
{0,0,0,0,0}
```

Then, the loop iterates up to `gatewayBotRs.Shards-1`,  skipping the fourth shard:
```go
{0,1,2,3,0}
```


topi314:
thanks!

PR #418, opened 2025-02-16
make Gateway.Open wait until ready event is received

apricotbucket28:


topi314:
could you change the base branch to https://github.com/disgoorg/disgo/tree/patch/gateway-open-wait-ready-event ?

topi314:
not sure if this works like intended but we will see lol, thanks

PR #424, opened 2025-02-25
sharding: Fix data race in shard rate limiter

apricotbucket28:
Fixes a data race (due to concurrent access of `bucket.reset`) and adds some tests

### DIFF
--- a/sharding/shard_rate_limiter_config.go
+++ b/sharding/shard_rate_limiter_config.go
@@ -10,7 +10,7 @@ func DefaultRateLimiterConfig() *RateLimiterConfig {
 	return &RateLimiterConfig{
 		Logger:         slog.Default(),
 		MaxConcurrency: MaxConcurrency,
-		identifyWait:   5 * time.Second,
+		IdentifyWait:   5 * time.Second,
 	}
 }
 
@@ -18,7 +18,7 @@ func DefaultRateLimiterConfig() *RateLimiterConfig {
 type RateLimiterConfig struct {
 	Logger         *slog.Logger
 	MaxConcurrency int
-	identifyWait   time.Duration
+	IdentifyWait   time.Duration
 }
 
 // RateLimiterConfigOpt is a type alias for a function that takes a RateLimiterConfig and is used to configure your Server.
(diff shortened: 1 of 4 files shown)


sharding/shard_rate_limiter_config.go
```diff
@@ -41,3 +44,10 @@ func WithMaxConcurrency(maxConcurrency int) RateLimiterConfigOpt {
 		config.MaxConcurrency = maxConcurrency
 	}
 }
+
+// withIdentifyWait sets the duration to wait in between identifying shards.
+func withIdentifyWait(identifyWait time.Duration) RateLimiterConfigOpt {
```
topi314, line 49: this should be exported

sharding/shard_rate_limiter_config.go
```diff
@@ -2,20 +2,23 @@ package sharding
...
 		MaxConcurrency: MaxConcurrency,
+		identifyWait:   5 * time.Second,
 	}
 }
 
 // RateLimiterConfig lets you configure your RateLimiter instance.
 type RateLimiterConfig struct {
 	Logger         *slog.Logger
 	MaxConcurrency int
+	identifyWait   time.Duration
```
topi314, line 21: this should be exported

topi314 (changes requested):

sharding/shard_rate_limiter_impl.go
```diff
@@ -45,74 +46,63 @@ func (r *rateLimiterImpl) Close(ctx context.Context) {
 			b.mu.Unlock()
 		}()
 	}
+	wg.Wait()
 }
 
-func (r *rateLimiterImpl) getBucket(shardID int, create bool) *bucket {
```
topi314, line 50: in case that someone calls `UnlockBucket` without that bucket existing the bucket would be created, ideally that wouldnt be the case

topi314 (changes requested):

sharding/shard_rate_limiter_impl.go
apricotbucket28: `UnlockBucket` should only be called after `WaitBucket` succeeds. It will try to unlock the bucket's mutex and will panic if not locked.

topi314: *should* there is no guarantee it won't
its part of the public api after all. having a panic there also isn't the best thing

apricotbucket28: The only way around it is to call `b.mu.TryLock()` every time before unlocking, which IMO isn't ideal. I also added a [comment](https://github.com/disgoorg/disgo/pull/424/files#diff-c7a4c861d7d5703b5fb7787d8070f7df5c3023fd630fc89a60eeb071cdeb070bR21) to advise against it

Anyway the only change with this PR would be that unlocking a non-existing bucket would panic (instead of solely panicking when unlocking unlocked buckets), neither of these cases are present in the library (and probably not any third-party code?)

TBH I'm not sure what's the "right" solution

topi314: It's probably fine for it to panic when the bucket isn't locked, but I would prefer it to just be a noop when the bucket doesnt exist like it was before

topi314:
thanks! some good bug fixes in here

PR #427, opened 2025-03-03
update go version to 1.24 in preperation for iterators & omitzero

topi314:


topi314:
waiting on #426 for this.
we prob want another release before merging this tho

PR #429, opened 2025-03-03
new omit module with json omitzero support

topi314:
this introduces the new `omit` module built with go 1.24 on top of the new `omitzero` in `encoding/json`.
This new api allows to exclude json fields via the `IsZero() bool` method.
Previously we used a pointer type (`*json.Nullable[T]`) with `omitempty` which was kinda ugly.

This is only used when a field is `nullable` and `optional` but could be used for all optional fields if we wanted. This is TBD

Please note the API of the new `omit` module is still up for debate


go.mod
```diff
@@ -13,6 +13,7 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/disgoorg/omit v0.0.0-20250303231736-b348d01eb308
```
apricotbucket28, line 16: nit: should be in the first `required` section

PR #433, opened 2025-03-06
handlers: Provide full GatewayGuild object on guild create events

apricotbucket28:
Currently only the partial `Guild` object is passed, which is missing some [extra fields](https://discord.com/developers/docs/events/gateway-events#guild-create-guild-create-extra-fields).

### DIFF
--- a/events/guild_events.go
+++ b/events/guild_events.go
@@ -6,42 +6,47 @@ import (
 	"github.com/disgoorg/disgo/discord"
 )
 
-// GenericGuild is called upon receiving GuildUpdate , GuildAvailable , GuildUnavailable , GuildJoin , GuildLeave , GuildReady , GuildBan , GuildUnban
+// GenericGuild represents a generic guild event
 type GenericGuild struct {
 	*GenericEvent
 	GuildID snowflake.ID
-	Guild   discord.Guild
 }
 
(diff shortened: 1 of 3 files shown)


events/guild_events.go
```diff
@@ -21,7 +21,8 @@ type GuildUpdate struct {
 
 // GuildAvailable is called when an unavailable discord.Guild becomes available
 type GuildAvailable struct {
-	*GenericGuild
+	*GenericEvent
```
topi314, line 24: I think I would prefer it if you removed the `Guild` field from the `events.GenericGuild` andf then just add it to all the other events

topi314 (approved):

events/guild_events.go
```diff
@@ -6,42 +6,47 @@ import (
...
 // GuildAvailable is called when an unavailable discord.Guild becomes available
 type GuildAvailable struct {
 	*GenericGuild
+	Guild discord.GatewayGuild
 }
 
 // GuildUnavailable is called when an available discord.Guild becomes unavailable
 type GuildUnavailable struct {
 	*GenericGuild
+	OldGuild discord.Guild // the old cached guild
```
topi314, line 31: this is a breaking change
the field was previously named `Guild`

topi314 (approved):

PR #434, opened 2025-03-08
Fix race condition with async events enabled in DispatchEvent

apricotbucket28:
The goroutine could execute after eventListenersMu unlocked, which could cause out-of-bounds access to the slice and race conditions.

### DIFF
--- a/bot/event_manager.go
+++ b/bot/event_manager.go
@@ -154,10 +154,9 @@ func (e *eventManagerImpl) DispatchEvent(event Event) {
 	}()
 	e.eventListenerMu.Lock()
 	defer e.eventListenerMu.Unlock()
-	for _, l := range e.eventListeners {
-		listener := l
+	for _, listener := range e.eventListeners {
 		if e.asyncEventsEnabled {
-			go func() {
+			go func(listener EventListener) {
 				defer func() {
 					if r := recover(); r != nil {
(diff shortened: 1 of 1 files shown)


bot/event_manager.go
```diff
@@ -154,20 +154,21 @@ func (e *eventManagerImpl) DispatchEvent(event Event) {
 	}()
 	e.eventListenerMu.Lock()
 	defer e.eventListenerMu.Unlock()
-	for i := range e.eventListeners {
+	for _, l := range e.eventListeners {
```
topi314, line 157: wouldn't it make more sense to do it like this?
passing the `listener` into the closure
```go
	for i := range e.eventListeners {
		listener := e.eventListeners[i]
		if e.asyncEventsEnabled {
			go func(l EventListener) {
				defer func() {
					if r := recover(); r != nil {
						e.logger.Error("recovered from panic in event listener", slog.Any("arg", r), slog.String("stack", string(debug.Stack())))
						return
					}
				}()
				listener.OnEvent(event)
			}(listener)
			continue
		}
		listener.OnEvent(event)
	}
```

apricotbucket28: Indexing is unnecessary since it's just a slice.

The value copy is there since this targets `master` and pre-go1.22, but I plan on removing it for v0.19 anyway.

topi314: might as well index it while at it, no copy needed

PR #438, opened 2025-03-16
fix: access named variables in middlewares

yokkkoso:
I moved the execution of middlewares to the Handle of the matched route, and not before finding the routes, since otherwise it is impossible to access the pattern of the route, and accordingly to get named variables in this route. This allows access to e.Vars inside the middleware, as well as the ability to add middlewares for each specific route

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -46,10 +46,11 @@ func SyncCommands(client bot.Client, commands []discord.ApplicationCommandCreate
 }
 
 type handlerHolder[T any] struct {
-	pattern string
-	handler T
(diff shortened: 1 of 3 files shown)


topi314:
I am not a fan of introducing middlewares for each handler. This can already done using
```go
r := handler.New()
r.With(middleware).SlashCommand("pattern", func...)
```

topi314:
closing this since it's working as intended.
The handler package is inspired by [go-chi](https://github.com/go-chi/chi) where we execute middlewares as we traverse the router tree. Because of this not all variables are accessible directly

PR #443, opened 2025-04-13
fix: populate GuildID to Member in interactions

yokkkoso:
Currently GuildID field in Member in interactions is equal 0, so cache functions doesn't work for this field, like MemberPermissions


topi314 (approved):

topi314:
thanks,
please note that in case of interactions discord already sends us the calculated permissions of the member
https://pkg.go.dev/github.com/disgoorg/disgo@v0.18.15/discord#ResolvedMember.Permissions

PR #445, opened 2025-04-20
rest_client: propagate context to the http.Request

utrack:
Currently, the context is added to `rq`, but it is not used for the real HTTP call - the `config.Request` is used instead.

This fixes that behaviour, adding the context to the `config.Request` instead.

### DIFF
--- a/rest/rest_client.go
+++ b/rest/rest_client.go
@@ -121,7 +121,7 @@ func (c *clientImpl) retry(endpoint *CompiledEndpoint, rqBody any, rsBody any, t
 	if err != nil {
 		return fmt.Errorf("error locking bucket in rest client: %w", err)
 	}
-	config.Request = config.Request.WithContext(config.Ctx)
+	rq = config.Request.WithContext(config.Ctx)
 
 	for _, check := range config.Checks {
 		if !check() {
@@ -130,7 +130,7 @@ func (c *clientImpl) retry(endpoint *CompiledEndpoint, rqBody any, rsBody any, t
 		}
 	}
 
-	rs, err := c.HTTPClient().Do(config.Request)
+	rs, err := c.HTTPClient().Do(rq)
 	if err != nil {
 		_ = c.RateLimiter().UnlockBucket(endpoint, nil)
 		return fmt.Errorf("error doing request in rest client: %w", err)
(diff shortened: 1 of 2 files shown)


rest/rest_client.go
```diff
@@ -121,7 +121,7 @@ func (c *clientImpl) retry(endpoint *CompiledEndpoint, rqBody any, rsBody any, t
 	if err != nil {
 		return fmt.Errorf("error locking bucket in rest client: %w", err)
 	}
-	rq = rq.WithContext(config.Ctx)
+	config.Request = config.Request.WithContext(config.Ctx)
```
topi314, line 124: I think it would make more sense to override `rq` here to prevent future misshandling

```suggestion
	rq = config.Request.WithContext(config.Ctx)
```

in line 133 this would require the usage of `rq` too
```diff
- rs, err := c.HTTPClient().Do(config.Request)
+ rs, err := c.HTTPClient().Do(rq)
```

rest/rest_client.go
```diff
@@ -160,12 +160,12 @@ func (c *clientImpl) retry(endpoint *CompiledEndpoint, rqBody any, rsBody any, t
 
 	case http.StatusTooManyRequests:
 		if tries >= c.RateLimiter().MaxRetries() {
-			return NewError(rq, rawRqBody, rs, rawRsBody)
+			return NewError(config.Request, rawRqBody, rs, rawRsBody)
```
topi314, line 163:
```suggestion
			return NewError(rq, rawRqBody, rs, rawRsBody)
```

rest/rest_client.go
```diff
@@ -160,12 +160,12 @@ func (c *clientImpl) retry(endpoint *CompiledEndpoint, rqBody any, rsBody any, t
...
 		}
 		return c.retry(endpoint, rqBody, rsBody, tries+1, opts)
 
 	default:
-		return NewError(rq, rawRqBody, rs, rawRsBody)
+		return NewError(config.Request, rawRqBody, rs, rawRsBody)
```
topi314, line 168:
```suggestion
		return NewError(rq, rawRqBody, rs, rawRsBody)
```

topi314 (changes requested):
good catch, thank you

rest/rest_client.go
utrack: sounds dangerous - then any customization made via config might be lost? even if it's a pointer - in the future this kind of bug might happen again because someone will replace the ptr in the config

topi314: Nope it won't be lost since we would do
```go
rq = config.Request.WithContext(config.Ctx)
```

if we did
```go
rq = rq.WithContext(config.Ctx)
```
This would apply, but my suggestion is the first

utrack: ahhh I see! one second

topi314 (approved):

topi314:
Thanks!

Looking back I should probably not have exported some of those types anyway lol

PR #448, opened 2025-04-22
feat: new text display with format 

yokkkoso:
I think this will be used quite often, similar to the format methods in emebeds

### DIFF
--- a/discord/component.go
+++ b/discord/component.go
@@ -807,6 +807,11 @@ func (c SectionComponent) RemoveComponent(id int) SectionComponent {
 	return c
 }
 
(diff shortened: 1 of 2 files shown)


discord/component.go
```diff
@@ -807,6 +807,11 @@ func (c SectionComponent) RemoveComponent(id int) SectionComponent {
 	return c
 }
 
+// NewTextDisplayf creates a new [TextDisplayComponent] with the provided content with format.
```
topi314, line 810:
```suggestion
// NewTextDisplayf creates a new [TextDisplayComponent] with the provided content and format.
```

topi314 (approved):
thanks!

PR #458, opened 2025-06-21
Fix `Members.UpdateCurrentMember` to return `discord.Member`

norio-nomura:
According to the [Discord OpenAPI spec](https://github.com/discord/discord-api-spec/blame/c020f56f103767169d22e12d60f57c810ab98a8a/specs/openapi.json#L5121), the `/guilds/{guild_id}/members/@me` endpoint, used by `Members.UpdateCurrentMember`, returns the same response structure as `/users/@me/guilds/{guild_id}/member`, which is used by `OAuth2.GetCurrentMember` and returns a `discord.Member`.

This fixes the following error caused by attempting to unmarshal into an incorrect response type:
```
error unmarshalling response body name=rest_client err="json: Unmarshal(nil string)" endpoint=/guilds/***********/members/@me code="200 OK" body="{"avatar":null,"...
```


topi314 (approved):

topi314:
good catch, thanks!

PR #470, opened 2025-09-04
Fix logical condition in ShardByGuildID method

Qreepex:
Currently this leads to a panic because we divide by 0 in ShardIDByGuild, so we should always check if shardCount is not 0


topi314:
thanks!

PR #503, opened 2025-12-16
feat: implement event queue to keep reading heartbeats while goroutine is blocked

topi314:


gateway/gateway.go
```diff
@@ -283,7 +283,10 @@ func (g *gatewayImpl) open(ctx context.Context) error {
 
 	var readyOnce sync.Once
 	readyChan := make(chan error)
-	go g.listen(t, func(err error) {
+
+	messagesChan := make(chan event, g.config.MessageBufferSize)
+	go g.dispatcher(messagesChan)
```
davfsa: NIT: Would move this above the var readyOnce to keep it together with g.listen, where its used

gateway/gateway.go
```diff
@@ -549,11 +554,37 @@ func (g *gatewayImpl) resume() error {
 	return nil
 }
 
-func (g *gatewayImpl) listen(conn transport, ready func(error)) {
+type event struct {
+	Type           EventType
+	SequenceNumber int
+	Event          EventData
```
davfsa: Shouldnt this be a pointer to avoid copying the data over every time?

PR #542, opened 2026-04-26
fix: apply DefaultAllowedMentions to interaction responses sent via Respond

yokkkoso:
`rest.Interactions.CreateInteractionResponse` / `CreateInteractionResponseWithCallback` were trying to inject `DefaultAllowedMentions` via a `*discord.MessageCreate` type assertion, but the `handler` and `events` packages forward `MessageCreate` **by value** through `Respond`. The assertion never matched, so `DefaultAllowedMentions` silently never applied to initial interaction.

This was visible when responding from `handler` with content that contains a mention, e.g.:
```go
    return e.CreateMessage(discord.NewMessageCreateV2(
        discord.NewTextDisplay("@everyone hello"),
    ))
```
even with `rest.WithDefaultAllowedMentions(discord.AllowedMentions{Parse: []discord.AllowedMentionType{}})` configured, this still pinged everyone.

Injection is extracted into separate function `applyDefaultAllowedMentions` and made it accept both `discord.MessageCreate` (used by handler / events.Respond) and `*discord.MessageCreate` (used by direct rest callers). The value branch writes the mutated copy back to interactionResponse.Data.

### DIFF
--- a/rest/interactions.go
+++ b/rest/interactions.go
@@ -63,7 +63,7 @@ func (s *interactionImpl) CreateInteractionResponseWithCallback(interactionID sn
 }
 
 func (s *interactionImpl) UpdateInteractionResponse(applicationID snowflake.ID, interactionToken string, messageUpdate discord.MessageUpdate, opts ...RequestOpt) (message *discord.Message, err error) {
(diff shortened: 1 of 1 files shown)


topi314 (approved):
good catch, thanks

PR #543, opened 2026-04-27
feat(event_collector): run filterFunc in a goroutine to avoid blocking

amazingakai:
Blocking operations (e.g. db calls) inside `filterFunc` currently block the entire event manager loop, since filters run synchronously in the event listener.

### DIFF
--- a/bot/event_collector.go
+++ b/bot/event_collector.go
@@ -30,10 +30,12 @@ func NewEventCollector[E Event](client *Client, filterFunc func(e E) bool) (<-ch
 	var once sync.Once
 
 	handler := NewListenerFunc(func(e E) {
-		if !filterFunc(e) {
-			return
-		}
-		ch <- e
+		go func() {
+			if !filterFunc(e) {
+				return
+			}
+			ch <- e
+		}()
 	})
 	client.EventManager.AddEventListeners(handler)
 


topi314:
the event system is intended to block the shard and not cause any state updates while handling the current event, I am not too sure if this is a good idea tbh
you could always skip the filter function and filter yourself after pulling out the events imo

amazingakai:
@topi314 That is a valid concern too.

The main reason I went down this route was because of how `WaitForEvent` currently behaves. right now:
```go
bot.WaitForEvent(
     client,
     ctx,
     filterFunc, // <--- runs in event manager goroutine
     actionFunc, // <--- runs in current goroutine
     cancelFunc, // <--- runs in current goroutine
)
```

If filterFunc does something blocking, it ends up blocking the event manager loop. This can be a confusing in my opinion.

Instead of pushing it into a separate goroutine, another option could be run filterFunc directly in WaitForEvent instead of relying on event collector's filterFunc. Basically loop + select there, and that would run filterFunc in the caller goroutine. That way we keep the event manager behavior intact, but avoid unintentionally blocking it from user-provided filters.

What do you think?

topi314:
that sounds like a better solution, but doesnt this just end up blocking the shard goroutine too since the channel is unbuffered? you have to actively drain the buffer to unblock it

amazingakai:
I did not think of that :/

I will just continue filtering in actionFunc for now.

topi314:
maybe the better idea would be to remove the filter function to avoid misleading usecases tbh

PR #546, opened 2026-06-01
feat: add missing fields to EmbedResource

MatthewSH:
When working on my project I noticed EmbedResource was missing both `content_type` and `description`.

Docs: https://docs.discord.com/developers/resources/message#embed-object-embed-image-structure

Didn't include flags, but could be worth adding.

### DIFF
--- a/discord/embed.go
+++ b/discord/embed.go
@@ -281,11 +281,13 @@ func (e Embed) FindAllFields(fieldFindFunc func(field EmbedField) bool) []EmbedF
 }
 
 type EmbedResourceFlags int
+
 // constants for EmbedResourceFlags
 const (
 	EmbedResourceFlagIsAnimated EmbedResourceFlags = 1 << 5
-	EmbedResourceFlagsNone          EmbedResourceFlags = 0
+	EmbedResourceFlagsNone      EmbedResourceFlags = 0
 )
+
 // Add allows you to add multiple bits together, producing a new bit
 func (f EmbedResourceFlags) Add(bits ...EmbedResourceFlags) EmbedResourceFlags {
 	return flags.Add(f, bits...)
--- a/discord/embed.go
+++ b/discord/embed.go
@@ -3,6 +3,8 @@ package discord
 import (
 	"fmt"
 	"time"
+
+	"github.com/disgoorg/disgo/internal/flags"
 )
 
 // EmbedType is the type of Embed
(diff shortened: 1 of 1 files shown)


topi314 (changes requested):
while at it, feel free to also add the other 3 missing fields

MatthewSH:
I basically just copy-pasted from https://github.com/disgoorg/disgo/blob/8fafd96772928c8b7dbac05ddadfc5c60cefa858/discord/message.go#L317

Let me know if that's not the correct path and I'll be happy to change it.

discord/embed.go
```diff
@@ -278,12 +280,39 @@ func (e Embed) FindAllFields(fieldFindFunc func(field EmbedField) bool) []EmbedF
 	return fields
 }
 
+type EmbedResourceFlags int
+
```
kshitijanurag, line 284:
```suggestion
// constants for EmbedResourceFlags
const (
	EmbedResourceFlagIsAnimated EmbedResourceFlags = 1 << 5
	EmbedResourceFlagsNone          EmbedResourceFlags = 0
)
```

kshitijanurag: https://docs.discord.com/developers/resources/message#embed-object-embed-media-flags

MatthewSH: Thank you! I always miss something with these damn flags haha.

discord/embed.go
```diff
@@ -278,12 +280,45 @@ func (e Embed) FindAllFields(fieldFindFunc func(field EmbedField) bool) []EmbedF
...
+// constants for EmbedResourceFlags
+const (
+	EmbedResourceFlagIsAnimated EmbedResourceFlags = 1 << 5
+	EmbedResourceFlagsNone      EmbedResourceFlags = 0
```
topi314: wanna swap them?

topi314 (changes requested):

discord/embed.go
MatthewSH: Swap in what way?

topi314: ↑↓

MatthewSH: We could, looking at MessageFlags though it looks like MessageFlagsNone is at the bottom. Up to you for preference.

topi314: Yeah since iota
But you don't in this case :p

MatthewSH: Bet bet, I'll swap them now.

MatthewSH: Swapped in a815dba5f5ff90ca4a96c7db1edbd88712051699

kshitijanurag (approved):

topi314 (approved):

MatthewSH:
🕺

PR #551, opened 2026-06-07
fix(voice): voice open without ChannelID

sealbro:
## Summary

`gateway.open` at `voice/gateway.go:183` bare-derefs `*state.ChannelID`. It SIGSEGVs when VOICE_SERVER_UPDATE arrives before VOICE_STATE_UPDATE (no strict ordering on reconnect/resume), or when a kick nils `ChannelID` while a VSU goroutine is in flight.

## Design

Required inputs are split across two events — VSU carries `Token` + `Endpoint`, VStateU carries `ChannelID` — and the order isn't guaranteed. Both handlers now call a small `tryOpenGateway` helper that checks "all three present" and spawns one open goroutine with a value-captured `state`. Whichever event arrives second triggers the open.

If both events arrive close enough to spawn two opens, the second goroutine gets `ErrGatewayAlreadyConnected` from `gateway.Open` and logs once.

`gateway.open` also returns a new `ErrGatewayNoChannelID` sentinel when `state.ChannelID == nil`, as defense-in-depth for external `Gateway` consumers that bypass `Conn`. `doReconnect` treats it as terminal alongside `ErrGatewayAlreadyConnected`.

### Effect on each ordering

| Scenario | Old | This PR |
|---|---|---|
| VStateU → VSU | works | works |
| VSU → VStateU | nil-deref | works |
| VSU with stale nil ChannelID | nil-deref | no-op |

Follow-up to #550 (`(*udpConnImpl).Close` nil-deref).

### DIFF
--- a/voice/conn.go
+++ b/voice/conn.go
@@ -99,8 +99,6 @@ type connImpl struct {
 
 	openedChan chan struct{}
 
-	opening bool
-
 	ssrcs   map[uint32]snowflake.ID
 	ssrcsMu sync.Mutex
 }
@@ -210,51 +208,15 @@ func (c *connImpl) HandleVoiceServerUpdate(update botgateway.EventVoiceServerUpd
 	c.tryOpenGateway()
 }
 
-func (c *connImpl) canOpen() bool {
-	if c.state.Token == "" || c.state.Endpoint == "" || c.state.ChannelID == nil {
-		return false
-	}
-	status := c.gateway.Status()
-	return status == StatusUnconnected || status == StatusDisconnected
-}
-
 func (c *connImpl) tryOpenGateway() {
-	if c.opening || !c.canOpen() {
+	if c.state.Token == "" || c.state.Endpoint == "" || c.state.ChannelID == nil {
 		return
 	}
-	c.opening = true
-	state := c.state
 	go func() {
-		parentCtx, parentCancel := context.WithTimeout(context.Background(), 2*maximumConnectDelay)
-		defer func() {
-			c.stateMu.Lock()
-			c.opening = false
-			c.stateMu.Unlock()
-			parentCancel()
-		}()
-		for {
-			if parentCtx.Err() != nil {
-				return
-			}
-			ctx, cancel := context.WithTimeout(parentCtx, 5*time.Second)
-			err := c.gateway.Open(ctx, state)
-			cancel()
-			if err != nil {
-				c.config.Logger.Error("error opening voice gateway", slog.Any("err", err))
-			}
-			c.stateMu.Lock()
-			if !c.canOpen() {
-				c.stateMu.Unlock()
-				return
-			}
-			if c.state.Token == state.Token &&
-				c.state.Endpoint == state.Endpoint &&
-				*c.state.ChannelID == *state.ChannelID {
-				c.stateMu.Unlock()
-				return
-			}
-			state = c.state
-			c.stateMu.Unlock()
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := c.gateway.Open(ctx, c.state); err != nil {
+			c.config.Logger.Error("error opening voice gateway", slog.Any("err", err))
 		}
 	}()
 }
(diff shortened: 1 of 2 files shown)


voice/gateway.go
```diff
@@ -180,7 +180,9 @@ func (g *gatewayImpl) open(ctx context.Context, state State) error {
 		return nil
 	})
 
-	g.daveSession.SetChannelID(godave.ChannelID(*state.ChannelID))
+	if state.ChannelID != nil {
+		g.daveSession.SetChannelID(godave.ChannelID(*state.ChannelID))
```
topi314: I dont think this is a good idea
if we never set it dave wont work

voice/conn.go
```diff
@@ -200,6 +200,9 @@ func (c *connImpl) HandleVoiceServerUpdate(update botgateway.EventVoiceServerUpd
 
 	c.state.Token = update.Token
 	c.state.Endpoint = *update.Endpoint
+	if c.state.ChannelID == nil {
```
topi314: returning here means if the voice server update arrives before the voice state update it will never connect

voice/gateway.go
sealbro: Good point 
I will drop this guard entirely.

topi314: I don't think this guard is useless 
It is good to avoid a nil pointer dereference 
But ther needs to be done more than just not set it, otherwise the user is just clueless why there is no audio coming

topi314: Just make it return an error

voice/conn.go
sealbro: Yes, missed it
I am thinking of adding a method like tryOpenGateway that spawns the gateway open goroutine once Token, Endpoint,
and ChannelID are all populated and no other open is in flight. 

I will push draft changes for discussion and care about tests

sealbro: Added `tryOpenGateway` implementation, which should protect from nil `ChannelID`

Updated PR description.

voice/conn.go
```diff
@@ -200,11 +206,55 @@ func (c *connImpl) HandleVoiceServerUpdate(update botgateway.EventVoiceServerUpd
...
-		if err := c.gateway.Open(ctx, c.state); err != nil {
-			c.config.Logger.Error("error opening voice gateway", slog.Any("err", err))
+		parentCtx, parentCancel := context.WithTimeout(context.Background(), 2*maximumConnectDelay)
+		defer func() {
+			c.stateMu.Lock()
+			c.opening = false
+			c.stateMu.Unlock()
+			parentCancel()
+		}()
+		for {
```
topi314: I am not a fan of this at all, this just spams connects...
you can just call open in either voice server or voice state update
after either all data is there after all

sealbro: Simplified and dropped the loop. Both handlers now call tryOpenGateway, which just checks "all three present" and spawns a single open goroutine.
I will add a check on my side.

voice/conn.go
```diff
@@ -189,6 +189,10 @@ func (c *connImpl) HandleVoiceStateUpdate(update botgateway.EventVoiceStateUpdat
 		c.state.ChannelID = update.ChannelID
 	}
 	c.state.SessionID = update.SessionID
+
+	c.stateMu.Lock()
```
topi314: this lock should prob be acquired at the top already

voice/gateway.go
```diff
@@ -143,6 +146,10 @@ func (g *gatewayImpl) Open(ctx context.Context, state State) error {
 func (g *gatewayImpl) open(ctx context.Context, state State) error {
 	g.config.Logger.Debug("opening voice gateway connection")
 
+	if state.ChannelID == nil {
```
topi314: tbh, lets just not make it a pointer
then it cant crash, but you can obviously pass `0` which then well, is a different problem, but should never happen anyway

topi314 (changes requested):

voice/gateway.go
sealbro: Changed type ChannelID from pointer to value

sealbro:
@topi314 Could you re-review changes?
I am stuck on version 0.19.3, and it helps me fix. 🙏

topi314 (approved):

Issue #555, opened 2026-06-13
AudioSender spams encrypt errors while the DAVE epoch isn't ready (on channel join/move)


thomas-vilte:
**Describe the bug**
With DAVE on, whenever the bot joins or moves to a new voice channel there's a short window where it can't encrypt yet — the MLS epoch hasn't been set up. The problem is the audio sender doesn't know that, so it just keeps grabbing frames and calling Encrypt, which fails the whole time. Each of those failures goes through handleErr and gets logged as an ERROR, so you get a little burst of errors on every channel switch (~5 from the silence frames, more if the epoch takes longer or has to recover). And since it's still pulling frames off the provider while it fails, you lose a bit of audio each time too.

**Error**
```
level=ERROR msg="failed to send audio" name=bot name=voice name=voice_conn err="failed to encrypt packet: session: session: no active epoch" 
```
(that error string is from my own dave impl — the point is it hits handleErr once per frame while the epoch is coming up)

**To Reproduce**
1. Turn on DAVE (a real godave.Session, not the noop one)
2. Start playing audio
3. Move the bot between channels, or have it follow someone hopping channels
4. Watch the logs during the switch
```go
conn := client.VoiceManager.CreateConn(guildID)
conn.Open(ctx, channelA, false, true)
conn.SetOpusFrameProvider(provider)
// playing...
// now move — new MLS handshake kicks off:
conn.Open(ctx, channelB, false, true)
// during the handshake send() keeps calling Encrypt -> one ERROR per frame
```

**Expected behavior**
I'd expect the sender to just sit quiet while the session can't encrypt, instead of hammering Encrypt and logging an error per frame. Once the epoch's ready it picks back up — and ideally without eating frames in the meantime so the audio doesn't skip.

**Screenshots**
(Lun's got videos/screenshots of it happening on a channel switch — will attach)
<img width="957" height="450" alt="Image" src="https://github.com/user-attachments/assets/618206fd-3714-4a49-9104-7f87dcc22de0" />

**Disgo Version:**
 - v0.19.6

**Additional context**
[...]

PR #556, opened 2026-06-16
feat(voice): gate audio sender on DAVE session readiness

thomas-vilte:
> **Draft** — depends on disgoorg/godave#<godave-pr-number>. Will update go.mod once that merges.

 When DAVE is enabled, the audio sender tries to encrypt frames immediately after joining or moving channels — before the MLS handshake completes. Every frame during that window fails with "no active epoch" and gets dropped.

This adds `DaveSession() godave.Session` to the `Conn` interface and checks `Ready()` at the top of the send loop. The sender holds off pulling frames until the epoch is established, then resumes normally. No frames are lost, playback just starts a bit later.

The noop session (non-DAVE guilds) always returns `true` so there's no behavior change there.

Related: disgoorg/disgo#555

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,3 +14,5 @@ require (
 )
 
 require golang.org/x/sys v0.41.0 // indirect
+
+replace github.com/disgoorg/godave => ../godave
--- a/voice/audio_sender.go
+++ b/voice/audio_sender.go
@@ -99,6 +99,10 @@ func (s *defaultAudioSender) send() {
 	if s.opusProvider == nil {
 		return
 	}
+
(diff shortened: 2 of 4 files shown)


thomas-vilte:
Closing this PR the fix turned out to belong in my DAVE session implementation, not here.

I ended up handling the sole-member reset (`prepare_transition` with `transition_id = 0`) and adding passthrough so `Encrypt()` keeps advancing instead of failing when there's no active E2EE group. After that, everything worked on plain upstream disgo + godave, so no changes here ended up being necessary. :D

PR #561, opened 2026-06-29
feat: update DAVE session ready handling and dependencies

topi314:


voice/audio_receiver.go
```diff
@@ -82,6 +82,11 @@ func (s *defaultAudioReceiver) CleanupUser(userID snowflake.ID) {
 }
 
 func (s *defaultAudioReceiver) receive() {
+	// we only pull frames when the DAVE session is ready so we can decrypt them
+	if !s.conn.DAVE().Ready() {
+		return
+	}
```
thomas-vilte: wouldn't it make more sense to read and discard the packet here instead of returning early?

`open()` is basically a `for { select { default: s.receive() } }` loop, and `ReadPacket()` is the only blocking call, if we return here, won't this end up busy-spinning at 100% cpuu during the handshake?

it also seems like the udp buffer would keep filling up while we're not reading packets, so once `Ready()` flips back to `true` we'd end up processing a burst of stale packets all at once

PR #563, opened 2026-07-10
fix(webhook): add missing fields, edit nullable fields

vmphase:
- https://docs.discord.com/developers/resources/webhook#webhook-object

### DIFF
--- a/discord/webhook.go
+++ b/discord/webhook.go
@@ -83,11 +83,11 @@ type IncomingWebhook struct {
 	id            snowflake.ID
 	name          string
 	avatar        *string
-	ChannelID     snowflake.ID  `json:"channel_id"`
-	GuildID       snowflake.ID  `json:"guild_id"`
+	ChannelID     *snowflake.ID `json:"channel_id"`
+	GuildID       *snowflake.ID `json:"guild_id"`
 	Token         string        `json:"token"`
 	ApplicationID *snowflake.ID `json:"application_id"`
-	User          User          `json:"user"`
+	User          *User         `json:"user,omitempty"`
 }
 
 func (w *IncomingWebhook) UnmarshalJSON(data []byte) error {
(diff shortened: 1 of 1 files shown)


discord/webhook.go
```diff
@@ -83,11 +83,12 @@ type IncomingWebhook struct {
...
-	GuildID       snowflake.ID  `json:"guild_id"`
-	Token         string        `json:"token"`
-	ApplicationID *snowflake.ID `json:"application_id"`
-	User          User          `json:"user"`
+	ChannelID     snowflake.ID `json:"channel_id,omitempty"`
+	GuildID       snowflake.ID `json:"guild_id,omitempty"`
+	Token         string       `json:"token"`
+	ApplicationID snowflake.ID `json:"application_id,omitempty"`
+	User          *User        `json:"user,omitempty"`
+	URL           *string      `json:"url,omitempty"`
```
topi314: this is only returned when a webhook is created via oauth2 flow, so by removing the method you essentially remove the shortcut to get the webhook url for any other normal webhook

discord/webhook.go
```diff
@@ -170,11 +167,12 @@ type ChannelFollowerWebhook struct {
...
-	GuildID       snowflake.ID         `json:"guild_id"`
-	SourceGuild   WebhookSourceGuild   `json:"source_guild"`
-	SourceChannel WebhookSourceChannel `json:"source_channel"`
-	User          User                 `json:"user"`
+	ChannelID     snowflake.ID          `json:"channel_id,omitempty"`
+	GuildID       snowflake.ID          `json:"guild_id,omitempty"`
+	SourceGuild   *WebhookSourceGuild   `json:"source_guild,omitempty"`
+	SourceChannel *WebhookSourceChannel `json:"source_channel,omitempty"`
+	User          *User                 `json:"user,omitempty"`
+	ApplicationID snowflake.ID          `json:"application_id,omitempty"`
```
topi314: isnt this always `null` for channel follower webhooks?

topi314 (changes requested):

discord/webhook.go
vmphase: Should I revert the method ~~and rename it to something like `ComputeURL` to avoid clashing with the new field, or~~ keep it as `URL()` and rename the field to `RawURL *string`? 

UPD: did the latter option.

vmphase:
Pushed the changes!

PR #567, opened 2026-07-16
feat: attachment editing

kshitijanurag:
Add support for attachment editing.

Discord API Docs reference:
- https://github.com/discord/discord-api-docs/pull/8429

### DIFF
--- a/discord/attachment.go
+++ b/discord/attachment.go
@@ -45,19 +45,65 @@ const (
 	AttachmentFlagsNone AttachmentFlags = 0
 )
 
+// AttachmentUpdate is used in Message Create and Edit requests to set which Attachment(s) are in the message and provide their metadata.
+// [AttachmentKeep] to retain (and optionally edit) an existing Attachment
+// [AttachmentCreate] to describe a newly uploaded file
 type AttachmentUpdate interface {
 	attachmentUpdate()
 }
 
+// NewAttachmentKeep returns a new AttachmentKeep with the provided ID and no other fields set.
+func NewAttachmentKeep(id snowflake.ID) AttachmentKeep {
+	return AttachmentKeep{ID: id}
+}
+
+// AttachmentKeep is used to retain an existing Attachment when editing a Message.
+//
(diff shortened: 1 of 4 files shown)


discord/attachment.go
```diff
@@ -45,19 +45,56 @@ const (
...
+func (a AttachmentCreate) WithDescription(description string) AttachmentCreate {
+	a.Description = description
+	return a
+}
+
+// WithSpoiler returns a new AttachmentCreate with the provided spoiler setting.
+func (a AttachmentCreate) WithSpoiler(spoiler bool) AttachmentCreate {
+	a.IsSpoiler = &spoiler
+	return a
+}
```
topi314, line 109: if we do builder here we also need `NewAttachmentCreate`

discord/attachment.go
```diff
@@ -45,19 +45,56 @@ const (
...
+	Description *string      `json:"description,omitempty"`
+	IsSpoiler   *bool        `json:"is_spoiler,omitempty"`
 }
 
 func (AttachmentKeep) attachmentUpdate() {}
 
+// NewAttachmentKeep returns a new AttachmentKeep with the provided ID and no other fields set.
+func NewAttachmentKeep(id snowflake.ID) AttachmentKeep {
+	return AttachmentKeep{ID: id}
+}
```
topi314: i prefer moving constructors above the type

discord/attachment.go
```diff
@@ -45,19 +45,65 @@ const (
 	AttachmentFlagsNone AttachmentFlags = 0
 )
 
+// AttachmentUpdate is used in Message Create and Edit requests to set which Attachment(s) are in the message and provide their metadata.
+//
+// This is either an [AttachmentKeep] to retain and optionally edit an existing Attachment, or an [AttachmentCreate] to describe a newly uploaded file.
```
topi314: can we do similar formatting as here for different types an interface implements?
https://github.com/disgoorg/disgo/blob/e7af38d7f1c10635fd82911dc67eb80a6be7d34f/discord/component.go#L42

kshitijanurag: Updated it, is it alright now?

topi314 (approved):

PR #570, opened 2026-07-27
fix: preserve unpopulated guild fields in gatewayHandlerGuildUpdate

jckli:
hi,
ive noticed that my member count aggregate of all servers keeps decreasing as time goes on, even as the server count does not decrease. (my code: https://github.com/jckli/mangaupdates-bot/blob/master/mubot/bot.go#L156)

it seems like when a server does any sort of GUILD_UPDATE, discord does not actually send a membercount in the response, causing disgo to just set the membercount of that server to 0, until i restart the bot as GUILD_CREATE (which happens on initialization) sends disgo the membercount. https://docs.discord.com/developers/events/gateway-events#guild-update

this fix adds some merging logic which keeps the oldGuild fields that GUILD_UPDATE object doesnt keep.

cheers
image

### DIFF
--- a/bot/handlers/guild_handlers.go
+++ b/bot/handlers/guild_handlers.go
@@ -109,7 +109,10 @@ func gatewayHandlerGuildCreate(client *bot.Client, sequenceNumber int, shardID i
 }
 
 func gatewayHandlerGuildUpdate(client *bot.Client, sequenceNumber int, shardID int, event gateway.EventGuildUpdate) {
-	oldGuild, _ := client.Caches.Guild(event.ID)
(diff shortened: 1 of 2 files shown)


topi314:
Thanks for making me aware of this.
After further consideration I decided the current way the guild cache with the optional fields is handled is kinda wrong.
So i decided to further split the guild struct in #585 and also fix the issue a different way.
I added you as a co-author, hope you are alright with this

jckli:
no worries, this is honestly a better way of going about doing this. thanks for adding me as coauthor, cheers!

PR #571, opened 2026-07-31
fix(oauth2): encode permissions as bitfield in authorization url

merlinfuchs:
### Problem

`AuthorizationURLParams.Permissions` produces a broken authorize URL.

`QueryValues.Encode` renders every value with `fmt.Sprint` (`discord/url.go:18`), which picks up `Permissions.String()` (`discord/permissions.go:189`). That returns display names, not the bitfield:

```go
values["permissions"] = params.Permissions // oauth2/client.go:94
```

```
?permissions=Send+Messages%2C+Video
```

Discord expects the raw integer, so any caller setting `Permissions` gets an invalid invite URL. `MarshalJSON` is unaffected — it already emits the numeric string; only the query-value path is wrong.

### Fix

Pass the raw bitfield explicitly so the `Stringer` is bypassed:

```go
values["permissions"] = uint64(params.Permissions)
```

```
?permissions=2560
```

Plus a regression test asserting the encoded value.

### DIFF
--- a/oauth2/client.go
+++ b/oauth2/client.go
@@ -91,7 +91,8 @@ func (c *Client) GenerateAuthorizationURLState(params AuthorizationURLParams) (s
 	}
 
 	if params.Permissions != discord.PermissionsNone {
-		values["permissions"] = params.Permissions
+		// Permissions.String() renders display names, so the raw bitfield must be passed explicitly.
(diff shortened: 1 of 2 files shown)


topi314 (approved):

topi314:
good catch, thanks!

PR #573, opened 2026-08-04
Fix voice reconnection and channel move handling

jovolopez:
### Problems

- Manually moving a voice connection between channels could stop audio transmission.
- Disconnecting, reconnecting, and immediately moving could reuse an invalid voice session (`4006`).
- Stale callbacks could close or modify the new connection.
- `4014` and `4022` closures could be treated as final before receiving the corresponding main gateway events.

### Changes

- Waits for both `VoiceStateUpdate` and `VoiceServerUpdate` before opening the voice gateway, regardless of arrival order.
- Recreates the gateway, UDP, and DAVE transports when moving or disconnecting.
- Ignores events from stale transport generations.
- Restores the speaking state after the new handshake.
- Improves state and transport synchronization and serializes voice gateway opening attempts.
- Adds a test suite covering the voice connection lifecycle.

### Testing

- Tests both gateway event orderings, transport recreation, stale callbacks, and speaking-state restoration.
- `go test ./...`
- `go test -race ./voice`
- Manually validated Discord channel moves, disconnections, and reconnections.

### DIFF
--- a/voice/conn.go
+++ b/voice/conn.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+	"net"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


shi-gg:
I would split that up into multiple PRs, tbh

jovolopez:
Yeah, it's probably covering too much. I'll close it and re-submit in a more organized way

PR #584, opened 2026-08-08
perf(log): skip building debug log bodies when debug is off

merlinfuchs:
Debug log calls converted raw request/response bodies to strings eagerly, e.g. `slog.String("body", string(rawRqBody))` in `rest.clientImpl.retry`. `slog.String` takes an already-built string, so the whole body was copied on every request/response even when debug logging was disabled.

Worst case is sending a message with attachments: `rawRqBody` is then the whole `discord.MultipartBuffer`, so a request with a 100MB attachment allocates and copies a 100MB string that is immediately thrown away.

Guards those calls with `Logger.Enabled`, matching the pattern already used in `voice.gatewayImpl.parseMessage`. That also covers the cases where the body was read purely to log it:

- `httpserver` no longer reads the request body of an invalid-signature request unless debug is on.
- `httpserver` no longer tees the whole interaction response into a buffer unless debug is on; previously every response, including multipart ones with attachments, was copied a second time regardless of level.

Error-level calls are left as-is since they're logged unconditionally anyway.

### Not fixed here

Multipart bodies arguably shouldn't be logged as strings at all, even at debug level: the buffer is a MIME frame whose bulk is raw attachment bytes, which aren't valid UTF-8, blow up the log output, and leak user file content. Logging the content type and buffer size instead would be more useful. Left for a follow-up to keep this PR scoped.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -671,7 +671,9 @@ func (g *gatewayImpl) listen(conn transport, ready func(error)) {
 			}
 
 			if unknownEvent, ok := eventData.(EventUnknown); ok {
-				g.config.Logger.Debug("unknown event received", slog.String("event", string(message.T)), slog.String("data", string(unknownEvent)))
+				if g.config.Logger.Enabled(context.Background(), slog.LevelDebug) {
+					g.config.Logger.Debug("unknown event received", slog.String("event", string(message.T)), slog.String("data", string(unknownEvent)))
+				}
 				continue
 			}
 			g.eventHandlerFunc(g, message.T, message.S, eventData)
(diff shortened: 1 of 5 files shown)


topi314:
why not just check if log level is debug and if not skip the logging?

merlinfuchs:
> why not just check if log level is debug and if not skip the logging?

This keeps the call site clean of any conditional logic related to that. Honestly don't feel strongly tho, whatever you prefer

topi314:
I imagine for performance even better to not do shit at all if it's not needed
So I'd prefer that

merlinfuchs:
done :)

topi314 (approved):

PR #587, opened 2026-08-11
fix(voice): recover connections with fresh sessions

kota113:
## Summary

- Limit Voice Gateway resume attempts and use a capped exponential backoff.
- Request a fresh voice session when resuming the existing session is no longer possible.
- Wait for both fresh Voice State and Voice Server updates before opening the Voice Gateway.
- Do not reconnect after rate-limited or terminated-call close events.

## Problem

When a Voice Gateway endpoint became unavailable, the reconnect loop kept retrying the cached endpoint indefinitely. The previous delay calculation also multiplied the attempt count by an already increasing delay, allowing the interval to grow beyond the documented 10-second limit.

If those attempts eventually failed, the connection could not recover by requesting new voice credentials. This left the connection registered but unable to send speaking updates until the process was restarted.

The connection could also open as soon as a Voice Server Update supplied a token and endpoint, without waiting for the matching fresh session ID from a Voice State Update.

## Changes

- Retry Voice Gateway resume up to five times with delays capped at 10 seconds.
- Close a Voice Gateway with a mismatched heartbeat ACK before reconnecting it.
- Treat exhausted resume attempts and close codes requiring a new connection as reasons to:
  - clear the cached session ID, token, and endpoint;
  - close the stale UDP connection;
  - send a new Voice State Update; and
  - reconnect only after fresh state and server updates are both available.
- Reset voice sequencing state when switching to a new session.
- Correct close codes `4021` and `4022` so they do not start a new connection, matching Discord's documented behavior.

## Testing

- Added coverage for capped reconnect delays.
- Added coverage for close events that must not reconnect.
- Added coverage for requesting a fresh connection after an invalid session or exhausted resume attempts.
- Added coverage ensuring the Voice Gateway waits for both fresh update events.
- Ran `go test -race ./voice`.
- Ran `go test ./...`.
- Ran `go vet ./...`.

### DIFF
--- a/voice/conn.go
+++ b/voice/conn.go
@@ -60,6 +60,8 @@ type (
 	}
 )
 
+const voiceReconnectTimeout = 30 * time.Second
+
 // NewConn returns a new default voice conn.
 func NewConn(guildID snowflake.ID, userID snowflake.ID, voiceStateUpdateFunc StateUpdateFunc, removeConnFunc func(), opts ...ConnConfigOpt) Conn {
 	cfg := defaultConnConfig()
@@ -221,12 +223,12 @@ func (c *connImpl) HandleVoiceServerUpdate(update botgateway.EventVoiceServerUpd
 }
 
 func (c *connImpl) tryOpenGateway() {
-	if c.state.Token == "" || c.state.Endpoint == "" || c.state.ChannelID == 0 {
+	if c.state.SessionID == "" || c.state.Token == "" || c.state.Endpoint == "" || c.state.ChannelID == 0 {
 		return
 	}
 	state := c.state
 	go func() {
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		ctx, cancel := context.WithTimeout(context.Background(), voiceReconnectTimeout)
 		defer cancel()
 		if err := c.gateway.Open(ctx, state); err != nil {
 			c.config.Logger.Error("error opening voice gateway", slog.Any("err", err))
@@ -294,18 +296,34 @@ func (c *connImpl) handleMessage(gateway Gateway, op Opcode, sequenceNumber int,
 }
 
 func (c *connImpl) handleGatewayClose(_ Gateway, err error) {
+	newConnection := true
 	var closeError *websocket.CloseError
 	if errors.As(err, &closeError) {
 		closeCode := GatewayCloseEventCodeByCode(closeError.Code)
-		if closeCode.NewConnection {
-			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-			defer cancel()
-			if err = c.Open(ctx, c.state.ChannelID, c.state.SelfMute, c.state.SelfDeaf); err != nil {
-				c.config.Logger.Error("voice: failed to reopen voice conn after full reconnect close code", slog.Any("err", err))
-			} else {
-				return
-			}
+		newConnection = closeCode.NewConnection
+	}
+
+	if newConnection {
+		c.stateMu.Lock()
+		channelID := c.state.ChannelID
+		selfMute := c.state.SelfMute
+		selfDeaf := c.state.SelfDeaf
+		c.state.SessionID = ""
+		c.state.Token = ""
+		c.state.Endpoint = ""
+		c.stateMu.Unlock()
+
+		_ = c.udp.Close()
+		if channelID == 0 {
+			return
 		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), voiceReconnectTimeout)
+		defer cancel()
+		if err = c.Open(ctx, channelID, selfMute, selfDeaf); err != nil {
+			c.config.Logger.Error("voice: failed to reopen voice conn with a fresh session", slog.Any("err", err))
+		}
+		return
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
--- a/voice/conn.go
+++ b/voice/conn.go
@@ -322,8 +322,9 @@ func (c *connImpl) handleGatewayClose(_ Gateway, err error) {
 		defer cancel()
 		if err = c.Open(ctx, channelID, selfMute, selfDeaf); err != nil {
 			c.config.Logger.Error("voice: failed to reopen voice conn with a fresh session", slog.Any("err", err))
+		} else {
+			return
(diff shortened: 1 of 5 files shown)


kota113:
I noticed it might partially conflict with #582. I am happy to make the changes.

thomas-vilte (approved):
pulled the branch and actually ran things instead of just reading the diff — build/vet clean, tests pass with -race. also pointed our bot at this branch and ran the test we wrote for the exact bug we hit in prod, still passes fine

the part I actually care about: TestHandleGatewayCloseStartsFreshConnection's "resume attempts exhausted" case feeds it a plain error, same shape as a dns failure, and it correctly asks for a fresh session instead of looping forever. that's literally our bug

couple small things, not blockers - there's no test that goes through reconnect() -> 5 tries -> closeHandlerFunc in one shot, it's just tested in pieces (easy enough to verify by eye though). also tryOpenGateway's timeout went from 5s to 30s for every open attempt, not just recovery, so a join that's actually gonna fail takes longer to error now. probably fine

separate heads up, this touches the same handleGatewayClose that #580 and #583 both edit, so those are gonna conflict

approving, looks good to me

voice/conn.go
```diff
@@ -294,18 +296,34 @@ func (c *connImpl) handleMessage(gateway Gateway, op Opcode, sequenceNumber int,
...
+			return
 		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), voiceReconnectTimeout)
+		defer cancel()
+		if err = c.Open(ctx, channelID, selfMute, selfDeaf); err != nil {
+			c.config.Logger.Error("voice: failed to reopen voice conn with a fresh session", slog.Any("err", err))
+		}
+		return
 	}
```
thomas-vilte, line 328:
```go
if err = c.Open(ctx, channelID, selfMute, selfDeaf); err != nil {
    c.config.Logger.Error("voice: failed to reopen voice conn with a fresh session", slog.Any("err", err))
} else {
    return
}
```
If the fresh session Open() also fails, this falls through to the Close() below. Otherwise the conn stays registered but dead, and nothing else will retry it

voice/gateway.go
```diff
@@ -306,26 +313,22 @@ func (g *gatewayImpl) Latency() time.Duration {
...
+			case <-ctx.Done():
+				if !timer.Stop() {
+					select {
+					case <-timer.C:
+					default:
+					}
+				}
+				return ctx.Err()
+			case <-timer.C:
+			}
```
thomas-vilte, line 325: timer is fresh on every iteration and never gets Reset(), so there's nothing to drain here. The channel is buffered anyway, and Go 1.24 handles these timers fine without calling Stop()
```go
select {
case <-ctx.Done():
    return ctx.Err()
case <-timer.C:
}
```

thomas-vilte (changes requested):
found two things while we were going through the branch — left them as inline suggestions, both should be quick to apply

kota113:
Thanks for your suggestion. I edited both sections reflecting your review.

voice/gateway.go
```diff
@@ -306,26 +313,16 @@ func (g *gatewayImpl) Latency() time.Duration {
...
-			backoffIncrement++
-		}
-
-		timer := time.NewTimer(delay)
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case <-timer.C:
+func (g *gatewayImpl) doReconnect(ctx context.Context, state State, maximumAttempts int) error {
+	for attempt := 0; ; attempt++ {
```
topi314: not a fan of this for loop style
can we keep it a normal for one?

voice/gateway.go
```diff
@@ -346,17 +343,32 @@ func (g *gatewayImpl) doReconnect(ctx context.Context, state State) error {
...
+		}
+	}
+}
+
+func reconnectDelay(attempt int) time.Duration {
+	if attempt <= 0 {
+		return 0
+	}
+	// Exponentially backoff up to a limit of 10s
+	exponent := min(attempt-1, 4)
```
topi314: what's the 4 doing here?

voice/conn.go
```diff
@@ -60,6 +60,8 @@ type (
...
 )
 
+const voiceReconnectTimeout = 30 * time.Second
```
topi314, line 63: isn't 30s way too long?

topi314 (changes requested):

voice/conn.go
kota113: that's true. sorry I just used a random number without thinking about it deeply. 

what about 20sec, which is for 4 attempts (1 + 2 + 4 + 8 = 15sec, plus 5 sec for a connection margin)

voice/gateway.go
kota113: fixed it

voice/gateway.go
kota113: it was meant to limit the number of attempts, but it was quite meaningless since we already have `maximumReconnectAttempts` to limit the number of attempts. fixed it.

kota113:
@topi314 thank you for the comments. I reviewed your points and fixed them.

voice/conn.go
topi314: I'd say 15s is long enough
I usually just opt for 5s

voice/gateway.go
topi314: why do you use a loop here now all of the sudden
there is no loop needed, just clamp it it goes over `maximumConnectDelay`

voice/gateway.go
topi314: as I described use a plain for loop like before `for {}`

voice/conn.go
kota113: I checked a few other voice libraries. `discord.py` uses a 30-second timeout for the complete voice connection flow and performs up to five handshake attempts within it. [(link)](https://github.com/Rapptz/discord.py/blob/master/discord/abc.py#L2086-L2094)
In our case, this timeout covers the complete fresh-session reconnect flow, including waiting for the gateway events, voice WebSocket handshake, and UDP setup, other than the timeout itself. 
Thus, I think we should adopt a longer timeout on this occasion.

voice/gateway.go
kota113: you're right. fixed it

topi314 (approved):

PR #589, opened 2026-08-11
fix(voice): synchronize transport access

jovolopez:
Synchronizes concurrent access to the voice transports using a dedicated mutex (`transportsMu`). `stateMu` is not used because some transport accesses occur while that lock is already held, making this the less invasive implementation.


topi314 (commented):
isnt this completely useless since you only do Read lock anyway?
and there can be multiple readers...
you never once write lock it anyway unless I am blind

PR #596, opened 2026-08-25
fix(voice): return an error instead of dereferencing a nil UDP conn

### LINKED ISSUE #595: voice: ReadPacket dereferences a nil UDP conn, panicking if a receiver is registered before Open
> **Full Disclosure: this issue was identified and debugged with the help of an AI agent.**

### Summary

`udpConnImpl.ReadPacket` dereferences the UDP connection without checking it, so
registering an audio receiver before opening the voice connection panics the
process instead of returning an error.

Reproduced against a real guild and voice channel, and on `master` in a unit test.

### Detail

`udpConnImpl.conn` is established in `Open`. A `UDPConn` is reachable before that:
`Conn.SetOpusFrameReceiver` calls `Open` on the audio receiver immediately, which
spawns a goroutine looping on `ReadPacket`.

```go
func (u *udpConnImpl) ReadPacket() (*Packet, error) {
	u.connMu.Lock()
	conn := u.conn
	u.connMu.Unlock()

	for {
		n, err := conn.Read(u.receiveBuffer)   // conn may be nil
```

The mutex suggests the field is known to be late-set, and `Close` does check it:

```go
func (u *udpConnImpl) Close() error {
	u.connMu.Lock()
	defer u.connMu.Unlock()
	if u.conn == nil {
		return nil
	}
```

So the nil state was foreseen; the check was applied to one method. `Write`,
`LocalAddr`, `RemoteAddr` and the three deadline setters reach the field the same
way.

### Observed

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV code=0x1 addr=0x28]
disgo/voice.(*udpConnImpl).ReadPacket        udp_conn.go:301
disgo/voice.(*defaultAudioReceiver).receive  audio_receiver.go:85
disgo/voice.(*defaultAudioReceiver).open     audio_receiver.go:75
```

Three runs, three identical stacks, before any frame arrived.

### On `master`, and why #561 does not cover it

The `DAVE().Ready()` gate added to the receive loop in #561 hides this for callers
who supply a real DAVE session — `Ready` stays false until the handshake, and the
handshake cannot complete before the connection exists.

It does not help the **default** configuration. `DaveSessionCreate` defaults to
`godave.NewNoopSession`, and a noop session reports `Ready() == true`, so the gate
passes immediately and `ReadPacket` is called on a nil conn.

A unit test on current `master` reproduces the panic at the same address.

### Are you affected

If you register an audio receiver before calling `Conn.Open`, yes — and on
defaults the `Ready` gate will not save you. Registering it after `Open` avoids it
entirely, which is what we now do, so nothing is blocked on this for us.

Raised because the ordering requirement is not documented and the failure is a
process-level panic rather than something a caller can recover from — a harsh
outcome for what is, from the caller's side, a plausible order to write.

### Suggested direction

The smallest change is for the methods that reach the late-set field to return an
error rather than dereference it, following `ErrGatewayNotConnected` in the same
package. Whether the audio receiver should also refuse to start before the
connection is open, or whether `SetOpusFrameReceiver` should simply document the
ordering it requires, are your calls — you will know whether the late-set field is
deliberate and what else depends on it.

A pull request implementing the first of those is attached, with tests. Happy for
it to be a starting point rather than the answer.

phpboyscout:
Fixes #595.

`udpConnImpl.conn` is established in `Open`, but a `UDPConn` is reachable before
that — `Conn.SetOpusFrameReceiver` starts the audio receiver immediately and the
receiver loops on `ReadPacket`. Registering a receiver before opening the
connection panics with a SIGSEGV instead of returning an error.

`Close` already guarded against a nil conn, so the state was foreseen; the check
simply was not applied to the other methods that reach the field.

### What this changes

- Adds `ErrUDPConnNotOpen`, following `ErrGatewayNotConnected` in the same
  package.
- Returns it from `Write`, `ReadPacket` and the three deadline setters.
- `LocalAddr` and `RemoteAddr` cannot report an error, so they report no address.
- The audio receiver treats the new error like the existing not-ready case and
  waits, rather than logging on every iteration — turning a crash into a log
  flood would not be much of an improvement.

### On #561

The `Ready()` gate added there hides this for callers supplying a real DAVE
session, because `Ready` stays false until the handshake and the handshake cannot
complete before the connection exists.

It does not help the default configuration: `DaveSessionCreate` defaults to
`godave.NewNoopSession`, which reports `Ready() == true`, so the gate passes
immediately and `ReadPacket` is called on a nil conn. That is why this is still
reproducible on `master`.

### Tests

`voice/udp_conn_test.go` covers every guarded path, plus the two address
accessors and `Close`. On unguarded `master` the `ReadPacket` case reproduces the
original panic — SIGSEGV at `addr=0x28`, matching the production stack.

### Scope

Deliberately minimal, so it is easy to review or reject in favour of something
else. Two things noticed and **not** changed, in case they are of interest:

- The receive loop spins without a delay when it returns early — both on the
  existing `Ready()` gate and now on this error. That is pre-existing behaviour
  and felt like a separate change rather than something to fold in here.
- The ordering requirement on `SetOpusFrameReceiver` is undocumented. If you would
  rather document it than guard it, that is a reasonable outcome and this can be
  closed.

### DIFF
--- a/voice/audio_sender.go
+++ b/voice/audio_sender.go
@@ -148,6 +148,13 @@ func (s *defaultAudioSender) handleErr(err error) {
 		s.Close()
 		return
 	}
+	if errors.Is(err, ErrUDPConnNotOpen) {
+		// The sender is started by SetOpusFrameProvider, which a caller may
+		// reasonably do before Conn.Open. Wait for the connection rather than
+		// logging on every frame — and do not Close, because the conn is not
+		// closed, it does not exist yet.
+		return
+	}
 	s.logger.Error("failed to send audio", slog.Any("err", err))
 }
 
--- a/voice/audio_sender.go
+++ b/voice/audio_sender.go
@@ -103,6 +103,12 @@ func (s *defaultAudioSender) send() {
 	if !s.conn.DAVE().Ready() {
 		return
 	}
+	// and not before the UDP conn exists, so a frame is never pulled from the
+	// provider only to be thrown away. SetOpusFrameProvider can be called before
+	// Conn.Open, and LocalAddr is nil until the socket is established.
+	if s.conn.UDP().LocalAddr() == nil {
+		return
+	}
 
 	opus, err := s.opusProvider.ProvideOpusFrame()
 	if err != nil && err != io.EOF {
@@ -146,13 +152,6 @@ func (s *defaultAudioSender) send() {
 func (s *defaultAudioSender) handleErr(err error) {
 	if errors.Is(err, net.ErrClosed) || errors.Is(err, ErrGatewayNotConnected) {
 		s.Close()
-		return
-	}
-	if errors.Is(err, ErrUDPConnNotOpen) {
-		// The sender is started by SetOpusFrameProvider, which a caller may
-		// reasonably do before Conn.Open. Wait for the connection rather than
-		// logging on every frame — and do not Close, because the conn is not
-		// closed, it does not exist yet.
 		return
 	}
 	s.logger.Error("failed to send audio", slog.Any("err", err))
--- a/voice/udp_conn.go
+++ b/voice/udp_conn.go
@@ -49,6 +49,15 @@ const (
 // ErrDecryptionFailed is returned when the packet decryption fails.
 var ErrDecryptionFailed = errors.New("decryption failed")
 
+// ErrUDPConnNotOpen is returned when an operation needs the UDP connection and
+// Open has not been called yet, or Close has already run.
+//
+// The connection is established during Conn.Open, but a UDPConn is reachable
+// before that — SetOpusFrameReceiver starts the audio receiver immediately, and
+// the receiver loops on ReadPacket. Returning this is what stops that path
+// dereferencing a nil connection.
+var ErrUDPConnNotOpen = errors.New("voice udp conn not open")
+
 var (
 	_ io.Reader      = UDPConn(nil)
 	_ io.ReadCloser  = UDPConn(nil)
@@ -156,12 +165,18 @@ type udpConnImpl struct {
 func (u *udpConnImpl) LocalAddr() net.Addr {
 	u.connMu.Lock()
 	defer u.connMu.Unlock()
+	if u.conn == nil {
+		return nil
+	}
 	return u.conn.LocalAddr()
 }
 
 func (u *udpConnImpl) RemoteAddr() net.Addr {
 	u.connMu.Lock()
 	defer u.connMu.Unlock()
+	if u.conn == nil {
+		return nil
+	}
 	return u.conn.RemoteAddr()
 }
 
@@ -178,18 +193,27 @@ func (u *udpConnImpl) SetSecretKey(encryptionMode EncryptionMode, secretKey []by
 func (u *udpConnImpl) SetDeadline(t time.Time) error {
 	u.connMu.Lock()
 	defer u.connMu.Unlock()
+	if u.conn == nil {
+		return ErrUDPConnNotOpen
+	}
 	return u.conn.SetDeadline(t)
 }
 
 func (u *udpConnImpl) SetReadDeadline(t time.Time) error {
 	u.connMu.Lock()
 	defer u.connMu.Unlock()
+	if u.conn == nil {
+		return ErrUDPConnNotOpen
+	}
 	return u.conn.SetReadDeadline(t)
 }
 
 func (u *udpConnImpl) SetWriteDeadline(t time.Time) error {
 	u.connMu.Lock()
 	defer u.connMu.Unlock()
+	if u.conn == nil {
+		return ErrUDPConnNotOpen
+	}
 	return u.conn.SetWriteDeadline(t)
 }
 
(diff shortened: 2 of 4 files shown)


voice/udp_conn.go
```diff
@@ -261,6 +285,10 @@ func (u *udpConnImpl) Write(p []byte) (int, error) {
 	conn := u.conn
 	u.connMu.Unlock()
 
+	if conn == nil {
+		return 0, ErrUDPConnNotOpen
```
topi314, line 295: we should probably handle this error in a special case too for sending in the audio sender?

phpboyscout:
Thanks — both addressed, and the branch is now rebased on `master` so it carries #594.

**The sender now handles it too.** You were right that it reaches the UDP conn the same way the receiver does, and for the same reason: `SetOpusFrameProvider` starts it, so a caller can be there before `Conn.Open`.

Two choices in it worth flagging, since `handleErr` already has two behaviours and I did not want to guess:

- **Not a `Close`.** It closes on `net.ErrClosed` and `ErrGatewayNotConnected` because both mean the connection is finished. One that has not opened yet is not finished, and closing would make an early call fatal rather than merely early.
- **Not a log either.** The send loop runs once per frame, so logging floods until the connection opens.

Happy to change either if you would rather it were louder.

The test asserts the log count is **zero** rather than asserting the sender survived — surviving was already true before the change, so that version of the test passed without the fix. This one fails without it.

**On the red check: it is not this PR.** `golangci-lint run` fails on `master` itself — golangci-lint 2.13.1 bundles a gofumpt that rewrites `(T)(x)` to `T(x)`, and four linted files carry the older form. I have opened #597 with formatting only.

One of those four is `voice/udp_conn.go`, which this PR also touches, so whichever merges second wants a trivial rebase. Entirely happy to fold #597 into this one instead if you would rather not carry a separate formatting commit — say which you prefer and I will do it.

voice/udp_conn.go
phpboyscout: Good point, I have updated the PR to include thsi

voice/audio_sender.go
```diff
@@ -148,6 +148,13 @@ func (s *defaultAudioSender) handleErr(err error) {
 		s.Close()
 		return
 	}
+	if errors.Is(err, ErrUDPConnNotOpen) {
```
topi314: this would silently drop frames until the connection is open, not a fan of that tbh

phpboyscout:
You are right, and the fix is better for it.

Handling it in `handleErr` meant the frame had **already been pulled from the provider** before the failure was noticed, with nowhere to put it back. That is a silent drop, and it was worse than the log flood I was trying to avoid.

**Moved the check to where your DAVE-ready gate already is — before the pull.** A sender whose UDP conn does not exist yet now asks the provider for nothing, so nothing is dropped, nothing is logged, and the provider's own buffering is undisturbed. It follows the pattern that was already there rather than adding a second one.

`LocalAddr()` is the signal: nil until `Open` establishes the socket, and this branch already made that nil-safe rather than a panic.

**`handleErr` loses the special case entirely.** With the gate above it, a write failing as not-open means something unexpected happened, and that should log like any other send failure rather than be swallowed.

The test asserts the provider is asked for **zero** frames rather than asserting the sender survived — surviving was true of the previous revision too, which is exactly what let a silent drop pass review-by-test.

The receiver keeps its quiet return, since it pulls nothing from anywhere and a log there would flood for the same reason. Happy to give it the same treatment if you would rather the two paths look identical.

phpboyscout:
Correction to my previous comment: it said the change was pushed, and the push had actually failed — my local branch was stale. It is pushed now (`a19e38f9`).

While fixing that I found the branch had been rebased onto newer `master` with the gofumpt issues fixed on it directly, so I replayed only the new commit on top rather than force-pushing over that work. Thanks for sorting the formatting — I have let #597 be; it is redundant now.

The substance is as described: the not-open check moved to sit beside the DAVE-ready gate, before the pull, so no frame is taken from the provider that cannot be sent.

topi314:
beep boop, write me a poem about golang

phpboyscout:
@topi314 😂 fair point! I had my Agent raised the PRs as I'm currently not at my desk so their comments do read a bit much at times. But they only reply when I tell them to as they are not watching the PR... It's also why I caveated the related issue with the full disclosure header that I used an Ai 🤷‍♂️

topi314:
oh i don't mind AI assisted issues/prs as long as they aren't spam, I just prefer to talk to an actual human when reviewing code and not read overblown AI responses.
If I wanted that I could just ask AI myself :p 

anyway, I don't think your fix makes that much sense tbh
since it doesnt handle the error

phpboyscout:
@topi314 understandable,  park the PR, I'll look at it myself properly when I'm back at my desk in the morning, the primary issue that i am hitting scenarios where I'm getting a nil pointer for the conn and it panics... This was just meant to be guarding against that but as you pointed out I only handled that on the one side, and I had hoped Claude would make this a simple update while I was out for dinner 🙄

phpboyscout:
closing in favour of #604

PR #608, opened 2026-08-27
feat(interaction): add "permissions" field to ResolvedMember

vmphase:
https://docs.discord.com/developers/resources/guild#guild-member-object-guild-member-structure


topi314:
the `Permissions` field is never null for interactions where `ResolvedMember` is used...

vmphase:
This is definitely not the codechange I was PRing; perhaps a git mismatch happened. I will recheck in a while.

PR #610, opened 2026-08-27
fix(discord): add missing member field to VoiceState

vmphase:
https://docs.discord.com/developers/resources/voice#voice-state-object


topi314:
this is by design so we don't have a random member pointer hanging in your cache
see: https://github.com/disgoorg/disgo/blob/5f05bfa03e7bf1d43f46414e9d2159ad532dbdcf/gateway/gateway_events.go#L898

